PR #12, opened 2019-05-13
KeyboardNavigation Component

eskil0312:


src/hooks/useKeyboardNavigation.ts
```diff
@@ -0,0 +1,54 @@
...
+    RIGHT: 39,
+    ENTER: 13,
+    ESCAPE: 27,
+};
+
+export type KeyBoardEvents = {
+    onUp?: Function,
+    onDown?: Function,
+    onLeft?: Function,
+    onRight?: Function,
```
MartinForre: Missing onEnter and onEscape?

src/hooks/useKeyboardNavigation.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export default (keyBoardEvents: KeyBoardEvents) => {
+    const [ref, setRef] = useState(null);
```
MartinForre: Would useState<HTMLElement | null>(null); prevent setting other random values as the ref?

src/hooks/useKeyboardNavigation.ts
```diff
@@ -0,0 +1,54 @@
...
+const KEY_HANDLERS = {
+    UP: "onUp",
+    DOWN: "onDown",
+    LEFT: "onLeft",
+    RIGHT: "onRight",
+    ENTER: "onEnter",
+    ESCAPE: "onEscape",
+};
+
+const KEY_CODES = {
```
MartinForre: Could KEY_HANDLERS and KEY_CODES be combined?
`
const KEY_HANDLERS = {
    38: "onUp",
};
`

and then

`
const handler = heyBoardEvents[KEY_HANDLERS[keyCode]];
`

PR #13, opened 2019-05-15
Tabs component

eskil0312:
Added Tabs and changed color for hover to frameless button


src/components/general/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,51 @@
+import * as React from "react";
+import * as PropTypes from "prop-types";
+import * as styles from "./styles.less";
+
+type TabsProps = {
+    onTabClick: (tabKey: string) => void,
```
MartinForre: Should this be onChange instead of onTabClick? Makes more sense if we're to support keyboard navigation etc.

src/components/general/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,51 @@
...
+    activeTabKey: string,
+    children: any,
+};
+
+const Tabs = ({ onTabClick, activeTabKey, children }: TabsProps) => {
+    const renderTabContent = () => {
+        const active = children.find(child => child.props.tabKey === activeTabKey);
```
MartinForre: If the children prop only contains one child it will not be an array (and therefore .find will throw an error). Use React.Children.toArray(children) to ensure an array.

src/components/general/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,51 @@
...
+import * as styles from "./styles.less";
+
+type TabsProps = {
+    onTabClick: (tabKey: string) => void,
+    activeTabKey: string,
+    children: any,
+};
+
+const Tabs = ({ onTabClick, activeTabKey, children }: TabsProps) => {
+    const renderTabContent = () => {
```
MartinForre: Declaring lots of functions within a functional component can be expensive and lead to memory leaks over time with lots of components and rerenders. Could this be extracted to a seperate component? e.g. `<TabContent activeTabKey={activeTabKey}>{children}</TabContent>`

src/components/general/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,51 @@
...
+        const active = children.find(child => child.props.tabKey === activeTabKey);
+        if (!active) {
+            return null;
+        }
+
+        return React.Children.map(active.props.children, child => React.cloneElement(child));
+    };
+    const renderTabPane = () => {
```
MartinForre: Same thing as renderTabContent. Extract this to a seperate component and then: `<TabPane onChange={onChange} activeTabKey={activeTabKey}>{children</TabPane>`

src/components/general/Tabs/Tab.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const tabClasses = classNames(styles.tab, {
+        [styles.current]: isCurrent,
+        [styles.hover]: isHovering && !disabled,
+        [styles.pressed]: isPressed && !disabled,
+        [styles.disabled]: disabled,
+    });
+    const titleClasses = classNames(styles.title);
+
+    return (
+        <button
```
MartinForre: Should we consider using anchor-tags instead of buttons? That way developers could provide a url/href property to Tab and we would redirect to that url. This would enencourage the use of url-navigation and allow users to open tabs in new... tabs

src/components/general/Tabs/styles.less
```diff
@@ -0,0 +1,77 @@
...
+
+.tabs{
+    width: 100%;
+    display: flex;
+    flex-direction: column;
+
+    .tabsPane {
+        display: flex;
+        flex-direction: row;
+        border-bottom: 1px solid #E6E6E6;
```
MartinForre: Will this color be part of the color tokens?

src/components/general/Tabs/styles.less
```diff
@@ -0,0 +1,77 @@
...
+
+            .title {
+                font-size: 16px;
+                padding: 0 calc(var(--grid-unit)* 2px) 0 calc(var(--grid-unit)* 2px);
+                user-select: none;
+                overflow: hidden;
+            }
+
+            &:focus {
+                outline: 1px dashed #BFBFBF;
```
MartinForre: Will this color be part of the color tokens?

src/components/general/Tabs/styles.less
```diff
@@ -0,0 +1,77 @@
...
+            }
+            &.disabled {
+                cursor: not-allowed;
+                color: #E6E6E6 !important;
```
MartinForre: Will this color be part of the color tokens?

src/components/general/Tabs/styles.less
```diff
@@ -0,0 +1,77 @@
...
+                outline:none;
+
+            }
+
+            &.hover {
+                background-color: fade(#666666, 5%);
```
MartinForre: Should this variant be part of the color tokens?

src/components/general/Tabs/styles.less
```diff
@@ -0,0 +1,77 @@
...
+            }
+
+            &.current {
+                color: #004E55;
```
MartinForre: Will this color be part of the color tokens?

src/components/general/Tabs/styles.less
```diff
@@ -0,0 +1,77 @@
...
+                border-bottom: 2px solid #004E55;
+                outline:none;
+                &.hover{
+                    background-color: initial;
+                    border-bottom: 2px solid @color-accent2;
+                    color: @color-accent2;
+                }
+                &.pressed {
+                    background-color: #DEEDEE;
+                    border-radius: calc(var(--grid-unit)* 1px) calc(var(--grid-unit)* 1px) 0px 0px;;
```
MartinForre: Duplicate semi-colon at the end

src/components/general/Tabs/styles.less
```diff
@@ -0,0 +1,77 @@
...
+                color: #004E55;
+                border-bottom: 2px solid #004E55;
+                outline:none;
+                &.hover{
+                    background-color: initial;
+                    border-bottom: 2px solid @color-accent2;
+                    color: @color-accent2;
+                }
+                &.pressed {
+                    background-color: #DEEDEE;
```
MartinForre: Will this color be part of the color tokens?

PR #18, opened 2019-05-27
Tabs scroll

eskil0312:
Added on tab click scroll to tabs component

### DIFF
--- a/src/components/general/Tabs/Tabs.tsx
+++ b/src/components/general/Tabs/Tabs.tsx
@@ -28,19 +28,21 @@ const TabContent: React.FC<TabContentType> = ({ children, activeTabKey }) => {
 
 const TabPane: React.FC<TabsProps> = ({ children, onChange, activeTabKey }) => {
     const tabsPaneRef = React.useRef<HTMLDivElement | null>(null);
+    const activeTabRef = React.useRef<HTMLElement | null>(null);
 
-    const scrollToTab = (tabRef: HTMLElement) => {
+    const scrollToTab = (tabRef: HTMLElement | null) => {
         if (!tabsPaneRef.current || !tabRef) {
             return;
         }
         const pane = tabsPaneRef.current;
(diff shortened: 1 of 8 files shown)


eskil0312:
Also change the keyboardnavigation to return the setRef function as well as the actual ref.

src/components/general/Tabs/Tabs.tsx
```diff
@@ -13,45 +13,64 @@ type TabContentType = {
...
+    };
+
     const clonedChildren = React.Children.map(children, child => {
         const { title, tabKey } = child.props;
         if (!title || !tabKey) {
             return null;
         }
         return React.cloneElement(child, {
             onChange: () => onChange(tabKey),
+            onCurrent: scrollToTab,
```
MartinForre: Couldn't this be solved with useEffect(scrollToTab, [activeTabKey]) ?

eskil0312: But how would you then get the ref to the active tab?

MartinForre: Set the ref here in the TabPane component and pass it along to the active child? 🤔

eskil0312: Can't figure out how to pass the ref to the child, React.clone element won't let me pass refs as it is a function component. Do you have any suggestion on how this can be done? Could React.forwardRef() solve this in any way?

MartinForre: forwardRef should allow you to accept refs, but I thought you should be able pass refs no matter what 🤔 Give it a try

eskil0312: Didn't manage to solve it with forward ref.Instead  passed the tab ref to the parent as a part of the onChange function. Could then use useEffect(scrollToTab, [activeTabKey]) to scroll to the active tab

PR #22, opened 2019-06-04
Error boundary and error message

eskil0312:


src/components/general/ErrorMessage/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+import styles from "./styles.less";
+import * as Icons from "./errorIcons";
+import classNames from "classnames";
+
+export enum ErrorTypes {
+    error= "error",
+    accessDenied= "accessDenied",
+    notFound= "notFound",
+    noData= "noData",
+    noAttachment= "noAttachment",
```
MartinForre: noAttachment and noActionCreated look like they're specific for meetings/reviews. We shouldn't have app specific error types and messages in a common component. Should be able to use the title and message props to provide that

src/components/general/ErrorMessage/styles.less
```diff
@@ -0,0 +1,49 @@
...
+
+    .messageContainer {
+        position: relative;
+        top: 0; left: 0; bottom: 0; right: 0;
+        display: flex;
+        align-content: center;
+        align-items: center;
+        justify-content: center;
+        flex-direction: column;
+        background: rgba(255, 255, 255, .8);
```
MartinForre: var(--colors-something)?

src/components/general/ErrorMessage/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+type ErrorMessageProps = {
+    hasError?: boolean,
+    errorType?:ErrorTypes,
+    message?: any,
+    resourceName?: string,
+    title?: string,
+    children?: any,
+    icon?: any,
+    button?: string,
```
MartinForre: Could this be something a bit more descriptive than "button"? "action" or something like that?

src/components/general/ErrorBoundary/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+      }
+
+    state = {
+        hasError: false,
+        retries: 0,
+    };
+
+    componentDidCatch(error, errorInfo) {
+
+        this.setState({ hasError: true });
```
MartinForre: Persist the error and errorInfo to provide more info to the error message?

src/components/general/ErrorMessage/index.tsx
eskil0312: Removed noAttachment and noActionCreated

src/components/general/ErrorMessage/styles.less
eskil0312: There was no background color was not in the design sheet so it was removed

src/components/general/ErrorMessage/index.tsx
eskil0312: Changed to action

src/components/general/ErrorBoundary/index.tsx
eskil0312: Now uses the error.message to provide more info. 
Is it necessary to provide more info than the message to the user?
All error information was used to provide to appInsight, but we don't have a implementation for that yet.

eskil0312:
The icons implementation will be replaced by the useIcon implementation in #21

src/components/general/ErrorBoundary/index.tsx
eskil0312: Added error and errorMessage on state for now

src/components/general/ErrorBoundary/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+            return message;
+        }
+        if (errorMessage !== "") {
+            return errorMessage;
+        }
+
+        return "Unhandled error message";
+    }
+
+    renderChildren() {
```
MartinForre: Is this used by anyone?

src/components/general/ErrorBoundary/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+        if (this.state.hasError) {
+            return null;
+        }
+
+        return this.props.children;
+    }
+
+    onButtonClick() {
```
MartinForre: Sould this be something more along the lines of "takeAction"? E.g takeAction = () => { if(this.state.hasError && this.props.onTakeAction) { return this.props.onTakeAction(); } window.location.reload(); }

The same would apply to the ErrorMessage component. Makes more sense to "Take action" when something goes wrong than to "button" 🙈

src/components/general/ErrorBoundary/index.tsx
eskil0312: Nope 😄

src/components/general/ErrorBoundary/index.tsx
eskil0312: Agreed, rewrote method and changed prop name from onButtonClick to onTakeAction

src/components/general/ErrorMessage/styles.less
```diff
@@ -9,7 +9,7 @@
 
     .messageContainer {
         position: relative;
-        top: 0; left: 0; bottom: 0; right: 0;
+        top: 20%; left: 0; bottom: 0; right: 0;
```
MartinForre: Does this have to be percentage?

eskil0312: A but unsure how to solve this. Discussed with Lotte and concluded that the regular error message should always have a 120 px margin from the top. The small message should always be centered. 

Could this be defined in the component like i did in the most recent update? Or should this be left to the user of the component

PR #23, opened 2019-06-11
Feature/pagination

janeirikforland:


src/components/data/pagination/components/NextPageButton/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+type NextPageButtonProps = {
+    onClick: () => void;
+    disabled: boolean;
+};
+
+const NextPageButton = ({ onClick, disabled }: NextPageButtonProps) => (
+    <button onClick={onClick} disabled={disabled}>
+        {'>'}
```
eskil0312: Should this be a icon?

src/components/data/pagination/components/PrevPageButton/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+type PrevPageButtonProps = {
+    onClick: () => void;
+    disabled: boolean;
+};
+
+const PrevPageButton = ({ onClick, disabled }: PrevPageButtonProps) => (
+    <button onClick={onClick} disabled={disabled}>
+        {'<'}
```
eskil0312: Should this be a icon?

src/components/data/pagination/components/PrevPageButton/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+type PrevPageButtonProps = {
+    onClick: () => void;
+    disabled: boolean;
+};
+
+const PrevPageButton = ({ onClick, disabled }: PrevPageButtonProps) => (
+    <button onClick={onClick} disabled={disabled}>
```
eskil0312: Could all these pagination buttons be made with our Button component?

eskil0312: I see now that we have those kind of buttons in the design sheet, but they are not made yet

src/components/data/pagination/components/PageButton/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+import styles from './styles.less';
+
+type PageButtonProps = {
+    pageIndex: number;
+    currentPageIndex: number;
+    onClick: EventHandler<SyntheticEvent>;
+};
+
+const PageButton = ({ pageIndex, currentPageIndex, onClick }: PageButtonProps) => (
+    <button className={currentPageIndex === pageIndex ? styles.current : null} onClick={onClick}>
```
MartinForre: Should the PageButton component accept a "isCurrent" option instead of figuring out if it's the current page itself?

src/components/data/pagination/components/RowsPerPage/index.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { Fragment } from 'react';
+
+const perPageOptions = [5, 10, 20, 50, 100];
```
MartinForre: Could this be a type? E.g. `type PerPageOptions = 5 | 10 | 20 | 50 | 100;` That way users of the pagination component (and DataTable or whatever uses the pagination internaly) would get intellisense on the available per page options and couldn't pass `543` or whatever as per per page

src/components/data/pagination/styles/index.less
```diff
@@ -0,0 +1,57 @@
...
+        cursor: pointer;
+        font-size: 14px;
+        box-sizing: border-box;
+
+        &:last-child {
+            margin-right: 0;
+        }
+
+        &:hover {
+            border-radius: .grid-unit(0.5px)[];
```
MartinForre: Since the border radius has nothing to do with the grid, we shouldn't use the grid unit to set border-radius. If we change the grid-unit in the future, the border-radius would change with it, instead of remaining at 4px.

PR #24, opened 2019-06-12
Dialog component

eskil0312:
Static dialog component. Needs the box-shadow component before it is finished

The story will be removed once the box shadow is in place


src/components/feedback/Dialog/styles.less
```diff
@@ -0,0 +1,34 @@
+@import '~styles/mixins.less';
+
+.container {
+    padding: .grid-unit(2px)[] 0 .grid-unit(1px)[] 0;
+    border: 1px solid var(--color-black-alt4);
```
MartinForre: This should probably be a color with an alpha? E.g. rgba(0, 0, 0, .2); or something. But since the dialogs should be controlled by the notification center, it might be better if the "scrim" is rendered by the notification center and not the dialog? Not sure about that one 🤔

src/components/feedback/Dialog/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+    return <div className={styles.container}>{children}</div>;
+};
+
+export const DialogTitle: React.FC<DialogProps> = ({ children }) => (
+    <div className={styles.dialogTitle}>{children}</div>
+);
+
+export const DialogContent: React.FC<DialogProps & DialogContentProps> = ({ children, scrollable }) => {
+    const contentClasses = classNames(styles.dialogContent, {
+        [styles.scrollable]: scrollable,
```
MartinForre: Is the scrollable option something that should be controlled by the user of the component, or by the component it self? You don't actually know that it should be scrollable until you have tested if the contents fits within the dialog. Would it be better to check the scrollHeight/Width within useEffect and apply the scrollable styling if the content overflows?

src/components/feedback/Dialog/styles.less
```diff
@@ -0,0 +1,34 @@
...
+    border-radius: 4px;
+    max-width: .grid-unit(35px)[];
```
MartinForre: Should the width and height be hardcoded instead of just max-width? Perhaps a fixed width and flexible height (min- and max-height) and then resort to scrollable when the content overflows the max-height?

src/components/feedback/Dialog/index.tsx
eskil0312: Discussed with Lotte, scroll is only used in special cases for mobile and is not needed for now. So I removed it.

src/components/feedback/Dialog/styles.less
eskil0312: I'm not sure if I understand, this is not scrim styling

MartinForre: My mistake. Misread border as background 🙈

PR #25, opened 2019-06-12
Initial snackbar component

eskil0312:
Snackbar component. Needs the box-shadow component before it is finished

The story will be removed once the box shadow is in place


src/components/feedback/Snackbar/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+}
+
+type SnackBarProps = {
+    verticalPosition?: verticalPositions;
+    horizontalPosition?: horizontalPositions;
+    open?: boolean;
+    autoHideDuration?: number;
+    onClose?: () => void;
+    message?: string;
+    actions?: Array<JSX.Element>;
```
MartinForre: According to the specification the snackbar allows 0-1 action(s). Should we, instead of accepting an array of elements, accept a single string?
E.g.
`
cancellable: true,
cancelLabel: "Undo",
onCancel: () => undoDelete();
`

src/components/feedback/Snackbar/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+};
+
+const SnackBar: React.FC<SnackBarProps> = ({
+    verticalPosition = verticalPositions.bottom,
+    horizontalPosition = horizontalPositions.left,
+    open,
+    autoHideDuration = 6000,
```
MartinForre: Instead of the snackbar component handeling the "auto hide" it should be handled by the notification center. It could however accept an abort signal from an [AbortController ](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)and hide when the signal is aborted.

src/components/feedback/Snackbar/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+export enum horizontalPositions {
+    left = 'left',
+    right = 'right',
+    center = 'center',
+}
+
+type SnackBarProps = {
+    verticalPosition?: verticalPositions;
+    horizontalPosition?: horizontalPositions;
+    open?: boolean;
```
MartinForre: Should the snackbar be open by default and then only close when the abort signal is aborted?

src/components/feedback/Snackbar/index.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from 'react';
+import styles from './styles.less';
+import classNames from 'classnames';
+import Button from '../../general/Button';
+import useElevationClassName from '../../../hooks/useElevationClassName';
+
+export enum verticalPositions {
```
MartinForre: Types and enums should be PascalCase, not camelCase

src/components/feedback/Snackbar/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+    top = 'top',
+    bottom = 'bottom',
+}
+export enum horizontalPositions {
```
MartinForre: Types and enums should be PascalCase, not camelCase

src/components/feedback/Snackbar/styles.less
```diff
@@ -0,0 +1,89 @@
...
+    --width-multiplier: 75px;
+    background-color: var(--color-black-alt1);
+    height: .grid-unit(6px)[];
+    width: calc(var(--width-multiplier)* var(--grid-unit));
+    border-radius: 4px;
+    display: flex;
+    align-items: center;
+    position: fixed;
+    justify-content: space-between;
+    transition: opacity .2s;
```
MartinForre: Have you discussed the transition effect with Lotte?

eskil0312: Yes

src/components/feedback/Snackbar/index.tsx
```diff
@@ -0,0 +1,80 @@
+import * as React from 'react';
+import styles from './styles.less';
+import classNames from 'classnames';
+import useElevationClassName from '../../../hooks/useElevationClassName';
+
+export enum VerticalPositions {
+    top = 'top',
```
MartinForre: Enum values should also be PascalCase 😜

eskil0312: 😃

PR #29, opened 2019-06-19
Banner component

eskil0312:
Simple banner component


src/components/feedback/Banner/index.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from 'react';
+import styles from './styles.less';
+import Button from '../../general/Button';
+import classNames from 'classnames';
+
+type BannerProps = {
+    message: string;
+    icon?: any;
+    cancelLabel: string;
```
MartinForre: The cancel button label should probably be "hard coded" within the component to keep it consistent.

PR #34, opened 2019-06-24
Text input component

eskil0312:


src/components/general/TextInput/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+            icon,
+            onBlur,
+            onIconAction,
+            ...props
+        },
+        ref
+    ) => {
+        const inputRef = React.useRef<HTMLInputElement>(null);
+        const [focus, setFocus] = React.useState(false);
+        const inputContainerRef = ref as React.RefObject<HTMLDivElement | null>;
```
MartinForre: As a user of this component I might expect the forwarded ref to point to the input field, not the wrapping container?

src/components/general/TextInput/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+        );
+
+        const inputIcon = icon && (
+            <div className={styles.icon} onClick={onIconAction}>
+                {icon}
+            </div>
+        );
+
+        const inputHelperText = React.useMemo(() => {
+            if (errorMessage && error) {
```
MartinForre: This would hide the required help text when an error is present 🤔 Is this by design?

eskil0312: This is by design

src/components/general/TextInput/index.tsx
eskil0312: Ref is now pointing to the input element

src/components/general/TextInput/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+    disabled?: boolean;
+    error?: boolean;
+    errorMessage?: string;
+    isOptional?: boolean;
+    placeholder?: string;
+    label?: string;
+    helperText?: string;
+    onChange: (newValue: string) => void;
+    value?: string;
+    icon?: any;
```
MartinForre: : ReactElement?

src/components/general/TextInput/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+            if (!error && !icon) {
+                return null;
+            }
+            const inputIcon = error ? (
+                <ErrorIcon outline={false} color="#FF3B3B" cursor="default" />
+            ) : (
+                icon
+            );
+            return (
+                <div className={styles.icon} onClick={onIconAction}>
```
MartinForre: Perhaps this could use the new IconButton component?

src/components/general/TextInput/index.tsx
eskil0312: Changed to ReactElement type

src/components/general/TextInput/index.tsx
eskil0312: The icon button is too big to be placed inside the text input component

PR #39, opened 2019-07-03
Selection controls

eskil0312:
Starting with radio button, switch and checkbox


src/components/general/SelectionControls/Checkbox/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+        const inputRef =
+            (ref as MutableRefObject<HTMLInputElement | null>) || useRef<HTMLInputElement | null>();
+
+        const containerClassNames = classNames(styles.checkboxContainer, {
+            [styles.disabled]: disabled,
+        });
+
+        useEffect(() => {
+            if (inputRef.current && indeterminate) {
+                inputRef.current.indeterminate = true;
```
MartinForre: This will never set indeterminate to false 🤔

PR #47, opened 2019-08-01
Dropdown component 

eskil0312:
Combined TextInput and Menu component to make Dropdown


src/components/general/Dropdown/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+    label?: string;
+    selections: DropdownSelections[];
+    onSelect?: (item: DropdownSelections) => void;
+    selected?: string;
+};
+
+const Dropdown: FC<DropdownProps> = ({ selections, label, onSelect, selected }) => {
+    const [open, setOpen] = useState(false);
+    const [inputValue, setInputValue] = useState('');
+    const [dropdownSelections, setDropdownSelections] = useState(selections);
```
MartinForre: Should we have a useEffect on the selections prop incase the selections changes? E.g. the possible options is fetched from an API and not available in the first render. And also: Should the selections prop be renamed to "options"? And the type to DropdownOption instead of DropdownSelections?

src/components/general/Dropdown/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+    const close = useCallback(() => open && setOpen(false), [open]);
+
+    const filterSearch = useCallback(inputValue => {
+        setDropdownSelections(
+            selections.filter(selection =>
+                selection.title.toLowerCase().includes(inputValue.toLowerCase())
+            )
+        );
+    }, []);
```
MartinForre: Should probably be a dependency on the selections/options prop here to rebuild the callback when they change

src/components/general/Dropdown/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+            <div
+                style={{
+                    position: 'absolute',
+                    width: rect.width,
+                    height: rect.height,
+                    top: rect.top,
+                    left: rect.left,
+                }}
+            >
+                <Router history={fusionContext.history}>
```
MartinForre: I think you can remove the <Router> and <FusionContext.Provider> in this case. It's in the overlay hook since a completely new render context is created

src/components/general/Dropdown/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+        return () => {
+            overlayContainer && overlayContainer.removeChild(element);
+        };
+    }, []);
+    if(show===false) {
+        return null;
+    }
+    return (
+        overlayContainer &&
```
MartinForre: Could combine this with if(show === false) to group the conditions for when nothing should be rendered. Easier for other developers to read and understand  the code

src/components/general/Dropdown/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    return (
+        <div className={styles.inputContainer} ref={inputContainerRef}>
+            <TextInput
+                onChange={value => {
+                    if (!open) {
+                        setOpen(true);
+                        return;
+                    }
+                    setInputValue(value);
+                    filterSearch(value);
```
MartinForre: I guess the correct way to perform the filter would be in a useEffect with dependency on the input value? 🤔

src/components/utils/RelativeOverlayPortal.tsx
```diff
@@ -0,0 +1,42 @@
...
+    relativeRef: MutableRefObject<HTMLElement | null>;
+    show?: boolean;
+};
+
+const RelativeOverlayPortal: FC<OverlayProps> = ({ children, relativeRef, show }) => {
+    const element = document.createElement('div');
+    const overlayContainer = useOverlayContainer();
+    const rect = useRelativePositioning(relativeRef);
+
+    useEffect(() => {
```
MartinForre: Is this neccecary at all? The `element` variable is not used elsewhere in the component. Either remove the `element` and `useEffect` or point the portal to the `element` instead of the `overlayContainer`

src/components/general/Dropdown/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+import {
+    TextInput,
+    DropdownArrow,
+    Menu,
+    useClickOutsideOverlayPortal,
+    RelativeOverlayPortal,
+} from '@equinor/fusion-components';
+import styles from './styles.less';
+
+type DropdownOption = {
```
MartinForre: Perhaps the `DropdownOption` type should have `selected?: boolean;` property to indicate which option is currently selected?

PR #48, opened 2019-08-08
Modal Side Sheet

eskil0312:
Starting with modal side sheet, regular side sheet will be implemented in a new branch


src/components/general/SideSheet/Modal/styles.less
```diff
@@ -0,0 +1,94 @@
...
+                .headerIcons{
+                    align-self: flex-end;
+                    display: flex;
+                    flex: 0;
+                }
+            }
+        }
+
+        &.comfortable {
+            --header-font-size: calc(var(--grid-unit) * 3px);
```
MartinForre: Are these the only changes between comfy and compact?

src/components/general/SideSheet/Modal/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    const containerClassName = classNames(styles.container, {
+        [styles.show]: isShowing,
+    });
+    const content = useMemo(() => {
+        if (!show) {
+            return null;
+        }
+
+        if (scrollable) {
```
MartinForre: Would this work if it was "scrollable" all the time? Could simplify the code and usage of the component quite a bit

src/components/general/SideSheet/Modal/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+                </div>
+            );
+        }
+
+        return <div className={styles.content}>{children}</div>;
+    }, [children, scrollable, show]);
+
+    return (
+        <Overlay show={show}>
+            <div className={containerClassName} onClick={close}>
```
MartinForre: Wasn't there a common Scrim component? 🤔

src/components/general/SideSheet/Modal/styles.less
```diff
@@ -0,0 +1,94 @@
...
+    opacity: 0;
+    transition: opacity .2s ease;
+    background-color: rgba(0, 0, 0, .32);
+
+    .modalSideSheet {
+        position: fixed;
+        top: 0; bottom: 0;
+        right: 0;
+        width: .grid-unit(80px)[];
+        transform: translateX(640px);
```
MartinForre: Use the grid unit here?

src/components/general/SideSheet/Modal/styles.less
```diff
@@ -0,0 +1,94 @@
+@import '~styles/mixins.less';
+@import '~styles/variables.less';
+@import '~styles/colors.less';
```
MartinForre: No need to import variables or colors. They're available as global variables within the fusion root component. In fact, they clutter the transpiled code by adding `:root { --grid-unit: 8; // etc. }` a bunch of times to the output and the DOM

src/components/general/SideSheet/Modal/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+                        </div>
+                    </header>
+                    {content}
+                </div>
+            </div>
+        </Overlay>
+    );
+};
+
+const Overlay = ({ children, show }) => {
```
MartinForre: Extract to common component? E.g. `OverlayPortal`

src/components/general/SideSheet/Modal/styles.less
eskil0312: I think so, but there is no drawings of compact mode so I'll have to discuss with Lotte

src/components/general/SideSheet/Modal/index.tsx
eskil0312: There was, was just removed for debugging. Re-added it

src/components/general/SideSheet/Modal/styles.less
eskil0312: Discussed with Lotte, these are the only changes between comf and comp

PR #52, opened 2019-08-12
Add string mask example to text input story

MartinForre:


src/hooks/useStringMask.ts
```diff
@@ -0,0 +1,120 @@
...
+export const applyStringMask = (mask: string, value: string) => {
+    const maskChars = mask.split('');
+    let valueIndex = 0;
+    const output: string[] = [];
+
+    for (let i = 0; i < maskChars.length; i++) {
+        const maskChar = maskChars[i];
+        const token = maskTokens.find(t => t.mask === maskChar);
+
+        if (valueIndex >= value.length) {
```
eskil0312: for loops in 2019? Can't this be done with map, reduce, forEach etc?

MartinForre: Probably. I'll have a look at it. Bit of brain farts in there seems like

MartinForre: @eskil0312 fixed

PR #56, opened 2019-08-15
Accordion

eskil0312:


src/components/general/Accordion/Accordion.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+            faucibus risus. Aenean maximus ipsum eros, eget facilisis enim rhoncus vel. Sed nec
+            venenatis nisi. Morbi feugiat auctor neque. Proin mauris mauris, maximus ut porta eget,
+            rhoncus ut dui. Pellentesque tempus metus non lacus hendrerit, quis tristique libero
+            feugiat. Quisque feugiat ultricies tempor. Cras vel diam gravida, laoreet mi facilisis,
+            blandit tortor.
+        </div>
+    );
+    return (
+        <div style={{ width: 500 }}>
+            <Accordion
```
MartinForre: Could we change this to be an Accordion component that contains all of the collapsible items? Like `<Accordion><AccordionItem isOpen={isOpenFirst} onChange={...} label="Accordion header">{accordionContent}</AccordionItem><AccordionItem isOpen={isOpenSecond} onChange={isOpenSecond} label="Right action" actionDirection="right">{accordionContent}</AccordionItem></Accordion>` 🤔

src/components/general/Accordion/styles.less
```diff
@@ -0,0 +1,69 @@
+@import '~styles/mixins.less';
```
MartinForre: comfortable/compact :)

src/components/general/Accordion/Accordion.stories.tsx
eskil0312: But would that main Accordion component contain any specific logic/styling or would it just return its children?

MartinForre: It would contain styling (how the items relate to each other) and _could_ contain logic if we need something like only one item open at the time

eskil0312: I see I see

PR #62, opened 2019-08-26
Org Chart

eskil0312:


src/components/org/OrgChart/hooks/useParentSize.ts
```diff
@@ -0,0 +1,29 @@
...
+export default (ref?: React.MutableRefObject<SVGElement | null>): [number, number] => {
+    const [size, setSize] = useState<Size>({ height: 0, width: 0 });
+
+    const handleResize = () => {
+        if (ref && ref.current && ref.current.parentElement) {
+            const parent = ref.current.parentElement;
+            setSize({ height: parent.offsetHeight, width: parent.offsetWidth });
+        }
+    };
+    useEffect(() => {
```
MartinForre: Here we could use useEventListener(window, 'resize', handleResize); or just a loop with requestAnimationFrame. The more event listeners we have on resize (especially with layout thrashing) the jankier the application gets

eskil0312: Now uses requestAnimationFrame

src/components/org/OrgChart/actions.ts
```diff
@@ -0,0 +1,105 @@
...
+        });
+        dispatch({
+            type: 'UPDATE_CENTER',
+            x: parentWidth / 2,
+            y: parentHeight / 2,
+        });
+    }, [parentHeight, parentWidth]);
+
+    useEffect(() => {
+        if (breadCrumbs) {
```
MartinForre: How does this handle breadcrumbs beeing "reset" to null etc?

eskil0312: Breadcrumbs can now be set to null

PR #69, opened 2019-08-29
Modified the accordion to also account for 'dynamic' content

janeirikforland:
Had to modify the accordion to account for dynamic content, that is content that takes up space after a certain amount of time.


src/components/general/Accordion/AccordionItem.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { FC, useRef, useState, useEffect } from 'react';
+import React, { FC, useRef, useState, useEffect, useCallback } from 'react';
```
eskil0312: Is useCallback in use?

janeirikforland: Nope. Will remove it

PR #70, opened 2019-08-29
added new icon

janeirikforland:
Added a new "Open in new" icon


src/components/icons/components/action/index.ts
```diff
@@ -7,3 +7,4 @@ export { default as CollapseIcon } from './CollapseIcon';
 export { default as CalendarIcon } from './CalendarIcon';
 export { default as InfoIcon } from './InfoIcon';
 export { default as HistoryIcon } from './HistoryIcon';
```
eskil0312: Should this also be exported from src/index.ts?

janeirikforland: It is ;)

PR #74, opened 2019-08-30
Fix slider percentage calculaction

MartinForre:


src/components/general/Slider/utils.ts
```diff
@@ -34,3 +34,5 @@ export const createMarkerFinder = (
         );
     };
 };
+
+// ((marker - full) / full) * 100
```
msalte: Can be removed

PR #84, opened 2019-09-03
Remove background color on slider

eskil0312:


src/components/general/Slider/styles.less
```diff
@@ -45,7 +45,6 @@
         top: 50%;
         transform: translateY(-50%);
         z-index: 2;
-        background: var(--color-primary);
```
MartinForre: Should probably apply similar changes to the markers aswell. Markers to the right of the dot are "highlighted"

eskil0312: Remove style and the logic for applying it or just the style?

MartinForre: Just style

eskil0312: Removed style

PR #99, opened 2019-09-18
Feature/sidesheet size

ingridklepsvik:


src/components/core/Header/styles.less
```diff
@@ -15,6 +16,12 @@
     right: 0;
     z-index: 2;
 
+    @media (max-width: @mobile-max-width) {
+        grid-template-areas: "start logo aside" "content content content";
+        grid-template-columns: max-content max-content max-content;
```
eskil0312: Should this be added to this branch?

src/components/general/SideSheet/Modal/index.tsx
```diff
@@ -10,21 +10,25 @@ import {
...
 
+export type SideSheetSize = 'xlarge' | 'large' | 'medium' | 'small';
+
 type ModalSideSheetProps = {
     children: ReactNode;
     header?: string;
     show?: boolean;
     onClose?: () => void;
     headerIcons?: ReactNode[];
+    size?: SideSheetSize;
```
eskil0312: Maybe we should default to xLarge so existing modal side sheets don't unexpectedly change size

src/components/core/Header/styles.less
ingridklepsvik: Removed it

src/components/general/SideSheet/Modal/index.tsx
ingridklepsvik: default to xLarge

src/components/core/Header/styles.less
```diff
@@ -1,4 +1,5 @@
 @import "~styles/mixins.less";
+@import '~styles/variables.less';
```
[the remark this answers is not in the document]
ingridklepsvik: No, its removed now

PR #101, opened 2019-09-19
Stepper and ArrowForwardIcon

msalte:
Added new component Stepper
Added new component ArrowForwardIcon


src/components/general/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,148 @@
...
+            position: i + 1,
+            disabled: c.props.disabled,
+        }));
+
+        setStepKeys(steps);
+
+        const startIndex = startStep && startStep > 0 ? startStep - 1 : 0;
+
+        setActiveStepKey(steps[startIndex].key);
+    }, [startStep]);
```
eskil0312: Should children also be in the useEffect dependency array?

PR #118, opened 2019-10-07
Add icon

GryNagel:
Adding add icon (+)


src/components/icons/components/content/index.ts
```diff
@@ -1,2 +1,3 @@
 export { default as BlockIcon } from './BlockIcon';
 export { default as SaveIcon } from './SaveIcon';
+export { default as AddIcon } from './AddIcon';
```
eskil0312: Remember to also add export from ".src/index.ts"

GryNagel: Thanks! Fixed!

PR #144, opened 2019-10-15
Nullable check for obs

eskil0312:


src/components/org/PositionCard/components/PositionInstance.tsx
```diff
@@ -29,8 +29,9 @@ const PositionInstanceComponent: React.FC<PositionInstanceProps> = ({
         instance && instance.assignedPerson ? instance.assignedPerson.name : 'TBN';
     const locationName =
         instance && instance.location && instance.location.name ? instance.location.name : 'TBN';
+    const obs = instance && instance.obs && instance.obs !== '' ? instance.obs : 'N/A';
 
-    const obsTooltipRef = useTooltipRef('OBS: ' + position.basePosition.name, 'below');
+    const obsTooltipRef = useTooltipRef('OBS: ' + obs, 'below');
```
msalte: Doesn't really matter but I prefer `OBS: ${obs}` for string concat.

msalte: Markdown parses ` as code.

PR #147, opened 2019-10-16
Feature/position card timeline

eskil0312:


src/components/org/PositionCard/components/TimelineInstance.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { useMemo } from 'react';
+import styles from '../styles.less';
+import { PositionInstance } from '@equinor/fusion';
+import { useTooltipRef } from '@equinor/fusion-components';
+import classNames from 'classnames';
+
+type TimelineInstanceProps = {
+    instance: PositionInstance;
+    currentInstance: PositionInstance | null;
```
MartinForre: What's the difference between `instance` and `currentInstance`?

src/components/org/PositionCard/components/TimelineInstance.tsx
```diff
@@ -0,0 +1,87 @@
...
+    calculator: (time: number) => number;
```
MartinForre: The position calculator is really an internal calculation that should belong to the TimelineInstance component, not the consumer

src/components/org/PositionCard/components/TimelineInstance.tsx
eskil0312: instance is the instance from the frominstances map(), currentInstance is the current

src/components/org/PositionCard/components/TimelineInstance.tsx
eskil0312: Moved calculator

src/components/org/PositionCard/components/TimelineInstance.tsx
MartinForre: Could the naming of these be changed to be less ambiguous?

eskil0312: Adde jsDoc

PR #158, opened 2019-10-24
Slider - elevated label

ingridklepsvik:
Added a prop elevated if you want to have the label on "top" of the slider

### DIFF
--- a/src/components/general/Slider/styles.less
+++ b/src/components/general/Slider/styles.less
@@ -103,11 +103,20 @@
             white-space: nowrap;
             font-size: 11px;
             cursor: pointer;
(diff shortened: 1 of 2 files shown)


src/components/general/Slider/styles.less
```diff
@@ -103,11 +103,20 @@
...
+         
         }
 
         &.isLowered {
            height: calc(var(--grid-unit) * 8px);
         }
+
+         &.isElevated {
+             label{
+                top: 0;
```
eskil0312: Maybe reset the bottom prop so the label does not end up filling the entire container

PR #163, opened 2019-10-31
Textarea component

ingridklepsvik:
Made a new textarea component. The width of the textarea can be controlled outside of the component. 
This textarea component will automatically resize as content changes.

### DIFF
--- a/src/components/general/TextArea/styles.less
+++ b/src/components/general/TextArea/styles.less
@@ -3,16 +3,18 @@
 .textAreaContainer {
    display: flex;
     flex-direction: column;
+    height: 100%;
 
     .textAreaContent {
         display: flex;
         justify-content: space-between;
-        min-height: .grid-unit(15px)[];
+        min-height: .grid-unit(10px)[];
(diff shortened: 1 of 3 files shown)


src/components/general/TextArea/styles.less
```diff
@@ -0,0 +1,145 @@
+@import '~styles/mixins.less';
+
+.textAreaContainer {
+   display: flex;
+    flex-direction: column;
+
+    .textAreaContent {
+        display: flex;
+        justify-content: space-between;
+        min-height: .grid-unit(15px)[];
```
grynag: Is it possible to remove the min height here? 120px is pretty large. If it needs a min height maybe the size of 2 rows of text could be the minimum? Or maybe take the min height or styling as a prop so that the user of the component can specify the minimum height?

ingridklepsvik: I have now changed the min height to the size of two rows of textinput

grynag: Great!

PR #164, opened 2019-10-31
New icons: mode, sync, more, arrowupward, arrowdownward. Fix add icon

grynag:
I have added four new icons:

- Sync
- More (three vertical dots)
- ArrowUpward
- ArrowDownward
- Image editicon story

The add icon had some weird spacing on the path, i re-exported the icon from Figma and added it again. It seems fine here.


src/components/icons/components/editor/ModeIcon.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+
+import useIcon, { IconProps } from '../../../../hooks/useIcon';
+
+const ModeIcon = (props: IconProps) => {
```
MartinForre: We allready have a pencil icon? It's called the EditIcon

grynag: My bad, didn't see the edit icon. Removed the Mode icon and added the edit icon so it's visible in Storybook.

PR #165, opened 2019-10-31
Remove max-width from searchable dropdown and textinput

grynag:
Remove max width from searchable drodown and textinput.

### DIFF
--- a/src/components/general/TextInput/TextInput.stories.tsx
+++ b/src/components/general/TextInput/TextInput.stories.tsx
@@ -45,7 +45,7 @@ const TextInputStory = () => {
     );
 
     return (
-        <div style={{ display: 'flex', padding: '8px', flexDirection: 'column', maxWidth: '328px' }}>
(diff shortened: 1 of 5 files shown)


src/components/general/TextInput/TextInput.stories.tsx
```diff
@@ -45,7 +45,7 @@ const TextInputStory = () => {
     );
 
     return (
-        <div style={{ display: 'flex', padding: '8px', flexDirection: 'column' }}>
+        <div style={{ display: 'flex', padding: '8px', flexDirection: 'column', maxWidth: '328px' }}>
```
MartinForre: Since we're removing the max-width of the text input, should we perhaps showcase that it _can_ grow to extreme widths?

grynag: Sure! I saw that the textarea was set up the same way, so it makes sense that they match. I've removed the max with from both the text input and the searchable dropdown stories.

PR #184, opened 2019-11-20
Expand sidesheet border to span entire width

sandves:
This is just a proposal. Just close it if it is not according to EDS.

**Before:**

image

**After:**

image


sandves:
Closing this as it was not according to EDS...

PR #187, opened 2019-11-25
Navigation drawer fix

ingridklepsvik:
Removed dropdown arrow when no children and section

### DIFF
--- a/src/components/general/NavigationDrawer/components/Grouping.tsx
+++ b/src/components/general/NavigationDrawer/components/Grouping.tsx
@@ -40,7 +40,7 @@ const Grouping: FC<NavigationComponentProps> = ({ navigationItem, onChange, isCo
                 </div>
             </>
         ),
-        [icon, title, isOpen, onChange]
+        [icon, title, isOpen, onChange, navigationChildren]
     );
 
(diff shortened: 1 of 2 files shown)


src/components/general/NavigationDrawer/components/Grouping.tsx
```diff
@@ -37,7 +34,9 @@ const Grouping: FC<NavigationComponentProps> = ({ navigationItem, onChange, isCo
                     className={styles.toggleOpenContainer}
                     onClick={() => onChange && onChange(id, true, false)}
                 >
-                    {!isCollapsed && <DropdownArrow cursor="pointer" isOpen={isOpen || false} />}
+                    {!isCollapsed && navigationChildren && navigationChildren.length > 0 && (
+                        <DropdownArrow cursor="pointer" isOpen={isOpen || false} />
```
eskil0312: Maybe add navigationChildren in the dependency array? If the structure gets updated/changed while the component is mounted, we also want the DropdownAroow to act accordingly

ingridklepsvik: Fixed

PR #188, opened 2019-11-26
ContextSelector rewrite to use standard dropdown-components.

olerichard:
Also changed to use contextHooks.


src/components/core/ContextSelector/index.tsx
```diff
@@ -1,177 +1,139 @@
...
-                        placeholder={'Search'}
-                        className={inputClassNames}
-                    />
-                    {isQuerying && <Spinner inline />}
-                </>
-            ) : (
-                <button className={styles.contextButton}>{getButtonContent()}</button>
-            )}
-        </div>
+        <HeaderContentPortal>
```
MartinForre: The context selector should _not_ controll where it's rendered. Thats up to the user of the component

src/components/core/ContextSelector/index.tsx
```diff
@@ -1,177 +1,139 @@
...
-        }
-
-        if (isFetchingRelatedContexts) {
-            return (
-                <span>
-                    Resolving {currentContext ? currentContext.title : null} <Spinner inline />
-                </span>
-            );
-        }
+    const updateOptions = (item: SearchableDropdownSection) =>
```
MartinForre: This does nothing? It maps the dropdown sections and returns it, but no one ever uses the return value

src/components/core/ContextSelector/index.tsx
```diff
@@ -1,177 +1,139 @@
...
+        }, [isOpen, queryText, selectedItem, currentContext]);
+
+        return (
+            <>
+                <SearchIcon color="#DADADA" />
+                <input
+                    type="text"
+                    value={selectedValue}
+                    onChange={e => setQueryText(e.target.value)}
+                    onClick={() => !isOpen && setIsOpen(true)}
```
MartinForre: We should use React.useCallback for these types of handlers to avoid recreating functions on each render and forcing the input component to rerender unnecessarily. We should also have a onKeyUp to catch the escape key and close the selector

src/components/core/ContextSelector/index.tsx
olerichard: I am unsure what this is.
Might be able to just remove it. It is a part of the ContextSelector in Pro-Org. 
Its just copy paste from there. 
Looks like it just does something in place and nothing else. 
Maybe there should be a call to setDropDownSections here.

You wrote this a couple months ago, maybe you remeber the intent here ?

MartinForre: Don't think i wrote this. I created the PR so the merge commit is by my user :P It's safe to remove.

src/components/core/ContextSelector/index.tsx
MartinForre: ☝️

src/components/core/ContextSelector/index.tsx
```diff
@@ -1,176 +1,146 @@
...
+            return '';
+        }, [isOpen, queryText, selectedItem, currentContext]);
+
+        const onChangeQueryText = React.useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
+            setQueryText(e.target.value);
+        }, []);
+
+        const onClickDropDown = React.useCallback(() => {
+            !isOpen && setIsOpen(true);
+        }, []);
```
MartinForre: Should have dependency on isOpen

src/components/core/ContextSelector/index.tsx
```diff
@@ -1,176 +1,146 @@
...
+                    onClick={onClickDropDown}
+                    onKeyUp={onKeyUpCloseDropDown}
+                    placeholder={selectedValue !== '' ? selectedValue : 'Search contexts'}
+                    className={styles.searchInput}
+                    ref={inputRef}
+                />
+                {isQuerying && <Spinner inline />}
+            </>
+        );
     }, []);
```
MartinForre: Sould have a dependency on isOpen, queryText, selectedItem and currentContext

olerichard: isOpen and selectedItem is  declared within the useDropDownController. Not sure how to get the hook to react on itself in that sense. 
As the isOpen and selectedItem get pulled off of the object at line 104. 
And i cant really push that up. 

Added queryText and currentContext as of now.

PR #191, opened 2019-11-29
Remove max-width of menu container

GryNagel:
### Issue
After updating the length on SearchableDropdown (https://github.com/equinor/fusion-components/commit/677c7c8720786d20d8793ccda7090a32caed6981), the selectable items in the dropdown menu does not fill the entire width:
image

The SearchableDropdown uses the Menu component, and that component has a max widht of `--grid-unit * 70px`. 

### Suggestion
This pull request removes the max width of the Menu component. **NOTE**: I don't know where this component is used and what it will affect, so this is a suggestion for a solution to the problem

### DIFF
--- a/src/components/general/Menu/__stories__/Menu.stories.tsx
+++ b/src/components/general/Menu/__stories__/Menu.stories.tsx
@@ -8,6 +8,27 @@ import { CheckBox, DoneIcon } from '@equinor/fusion-components';
 const MenuStory = () => {
     const [ref, setRef] = React.useState<HTMLElement | null>(null);
 
+    const items = [
+        {
+            key: '1',
+            title: 'First',
+        },
+        {
+            key: '2',
+            title: 'Selected',
+            isSelected: true,
+        },
+        {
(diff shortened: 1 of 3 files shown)


MartinForre:
Please update the storybook with examples on how to "control" the width of the menu component, as well as the searchable dropdown

src/components/general/Menu/__stories__/Menu.stories.tsx
```diff
@@ -17,29 +38,23 @@ const MenuStory = () => {
...
-                            {
-                                key: '4',
-                                title: 'Last',
-                            },
-                        ],
+                        items: items,
                     },
                 ]}
             />
+            <div style={{width: '300px'}}>
```
GryNagel: I've added a "Narrow version" of the drop down here. I haven't seen a lot of text explaining how or why in the Storybook so it's just the component wrapped in a div with a set width. What do you think @MartinForre? I've done the same thing with the menu component 😄

PR #195, opened 2019-12-04
Context Selector - Simplified context-model

olerichard:
Fixed Stories to work with new package.
Fixed bug where selectedValue for dropdown didnt update properly.
It used to set previously selected context instead of the one just selected.

### DIFF
--- a/src/components/core/ContextSelector/index.tsx
+++ b/src/components/core/ContextSelector/index.tsx
@@ -45,10 +45,6 @@ const ContextSelector: React.FC = () => {
         search(queryText);
     }, [queryText]);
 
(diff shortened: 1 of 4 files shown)


src/components/core/ContextSelector/index.tsx
```diff
@@ -103,7 +99,7 @@ const ContextSelector: React.FC = () => {
                 </>
             );
         },
-        [queryText, currentContext]
+        [queryText, currentContext, dropdownSections]
```
olerichard: This fixes the problem with the item selection. 
causing dropdownsections to update so selectedItem dosnt run with memoized sections at line 57

PR #196, opened 2019-12-04
Modale side sheet safe close

eskil0312:
Added a safeClose prop to the modal side sheet. It will notify the user that the closing was intentional before actually closing it

### DIFF
--- a/src/components/general/SideSheet/Modal/index.tsx
+++ b/src/components/general/SideSheet/Modal/index.tsx
@@ -50,9 +50,9 @@ export default ({
         if (response.confirmed || response.dismissed) {
             setIsShowing(false);
         }
-    }, []);
+    }, [sendNotification]);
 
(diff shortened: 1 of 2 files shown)


src/components/general/SideSheet/Modal/index.tsx
```diff
@@ -28,18 +29,35 @@ export default ({
...
     useEffect(() => {
         if (!isShowing && show) {
             setIsShowing(true);
         }
     }, [show]);
 
+    const closeSafe = useCallback(async () => {
+        const response = await sendNotification({
+            level: 'high',
+            title: 'Are you sure you want to close, all changes will be lost',
```
MartinForre: Should the title be parameterized?

eskil0312: The title is now a optional prop

Issue #198, opened 2019-12-09
Searchable dropdown has no scroll


tony-hagen:
If the dropdown contains a lot of items, the dropdown extends beyond the screen and is not scrollable.

eskil0312:
Merged and published to npm. Use prop `dropDownMaxHeight` to select max height. Content that overflows will scroll.

PR #199, opened 2019-12-09
Max height prop for dropdown

eskil0312:


eskil0312:
Ref #198

src/components/general/SearchableDropdown/styles.less
```diff
@@ -1,10 +1,7 @@
 @import '~styles/mixins.less';
 
 .menuContainer{
```
eskil0312: .menuContainer was not used until now

PR #200, opened 2019-12-11
SelectedPerson should be handled outside PersonPicker

sandves:
# Motivation
Currently, it is not possible to clear the selected person in the `PersonPicker` component.

# Solution
To solve this, `selectedPerson` is now passed in as a prop and should be handled by the parent component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@equinor/fusion-components",
-    "version": "0.1.78",
+    "version": "0.1.79",
     "description": "Common react components used by fusion core and fusion apps",
     "main": "dist/index.js",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@equinor/fusion-components",
-    "version": "0.1.78",
+    "version": "0.1.79",
```
MartinForre: We shouldn't bump version in pull requests. Can result in confusions if multiplie PRs bumps to different versions and are merged in the wrong order

eskil0312: Could we do the versioning in master?

MartinForre: Yes

PR #201, opened 2019-12-11
Resizable sidesheet

MartinForre:


src/components/general/SideSheet/Modal/index.tsx
```diff
@@ -31,7 +33,11 @@ export default ({
     headerIcons,
     size = 'xlarge',
```
eskil0312: Considering we have changed the sizes, should we set the default to 'large' instead of 'xlarge' so existing implementations don't change size unexpectedly?

src/components/general/SideSheet/Modal/styles.less
```diff
@@ -84,6 +90,46 @@
...
+        left: calc(var(--grid-unit) * -.5px);
+        cursor: col-resize;
+    
+        .indicator {
+            position: absolute;
+            top: 50%;
+            left: calc(var(--grid-unit) * -1px);
+            width: calc(var(--grid-unit) * 3px);
+            height: calc(var(--grid-unit) * 3px);
+            background: white;
```
eskil0312: var(--color-white) from color palette?

src/components/general/SideSheet/Standard/styles.less
```diff
@@ -77,5 +85,51 @@
...
+        left: calc(var(--grid-unit) * -.5px);
+        cursor: col-resize;
+    
+        .indicator {
+            position: absolute;
+            top: 50%;
+            left: calc(var(--grid-unit) * -1px);
+            width: calc(var(--grid-unit) * 3px);
+            height: calc(var(--grid-unit) * 3px);
+            background: white;
```
eskil0312: var(--color-white)?

src/components/general/SideSheet/Modal/index.tsx
MartinForre: Good idea. Fixed

PR #203, opened 2019-12-12
Fix standard sidesheet 100% width on mobil and resize handle on hover

MartinForre:


src/components/general/SideSheet/Standard/index.tsx
```diff
@@ -65,6 +65,7 @@ const SideSheet: React.FC<StandardSideSheetProps> = ({
         <div className={containerClassNames} key={id} style={{ ...resizedSize }}>
             {isResizable && isOpen && (
                 <div className={styles.resizeHandle} onMouseDown={onResizeStart}>
```
eskil0312: Is resizeHandle jsx and css the same for Modal and Standard side sheet? How about moving it to a shared component?

MartinForre: It's the same markup, but the styling may differ and also depends on styling based on the parent container

eskil0312: Alright

PR #205, opened 2019-12-13
Disc skeleton

ingridklepsvik:


src/components/feedback/Skeleton/styles.less
```diff
@@ -14,6 +14,62 @@
     height: calc(var(--grid-unit) * 4px);
 }
 
+.disc{
+    border-radius: 50%;
+    display: flex;
+    position: relative;
+    width: var(--photo-size);
```
MartinForre: Rename this variable to `--disc-size`?

src/components/people/PersonCard/index.tsx
```diff
@@ -70,6 +71,21 @@ export default ({
             {currentPerson && (
                 <div className={containerClassNames}>
                     <PersonPhoto person={currentPerson} size={photoSize} hidePopover />
+                    {isFetching && (
```
MartinForre: If I understand this code correctly, the skeleton will never show since `isFetching `will always be false when `currentPerson `is set?

src/components/people/PersonCard/index.tsx
```diff
@@ -32,9 +33,9 @@ export default ({
     showJobTitle,
 }: PersonCardProps) => {
     const [currentPerson, setCurrentPerson] = useState<PersonDetails>();
-    const { error, personDetails } = personId
+    const { isFetching, error, personDetails } = personId
         ? usePersonDetails(personId)
-        : { error: null, personDetails: person };
+        : { isFetching: true, error: null, personDetails: person };
```
MartinForre: We should probably have isFetching as an optional prop (isFetchingPerson?: boolean) and use this as isFetching when using the person prop instead of the personId prop

src/components/feedback/Skeleton/styles.less
ingridklepsvik: Renamed

src/components/people/PersonCard/index.tsx
ingridklepsvik: Fixed

src/components/people/PersonCard/index.tsx
```diff
@@ -30,11 +38,12 @@ export default ({
     photoSize = 'xlarge',
     hidePopover,
     showJobTitle,
+    isFetchingPerson,
 }: PersonCardProps) => {
     const [currentPerson, setCurrentPerson] = useState<PersonDetails>();
-    const { error, personDetails } = personId
+    const { isFetching, error, personDetails } = personId
```
eskil0312: This condition will cause a inconsistent hook call, usePersonDetails will only be called if personId is set. This is against the [rules of hooks](https://reactjs.org/docs/hooks-rules.html )

src/components/feedback/Skeleton/Disc.tsx
```diff
@@ -0,0 +1,50 @@
...
+        width: sizes[size],
+        height: sizes[size],
+        ...{
+            viewBox: `0 0 ${sizes.xlarge} ${sizes.xlarge} `,
+        },
+    };
+
+    return (
+        <div className={discClassNames}>
+            <FallbackIcon {...iconProps} />
```
MartinForre: The fallback icon belongs to the person photo component and has nothing to do with a skeleton

ingridklepsvik: removed it

src/components/feedback/Skeleton/Bar.tsx
```diff
@@ -5,17 +5,18 @@ import classNames from 'classnames';
 
 type SkeletonBarProps = {
     width?: number | string;
+    height?: number | string;
 };
 
-const SkeletonBar: React.FC<SkeletonBarProps> = props => {
+const SkeletonBar: React.FC<SkeletonBarProps> = ({ height, ...props }) => {
```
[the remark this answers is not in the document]
ingridklepsvik: Destructed both

src/components/people/PersonCard/index.tsx
eskil0312: This behavior is already in master and can be fixed in a separate pr

PR #206, opened 2019-12-17
Sidesheet initial resizes on mouse move

MartinForre:
## Problem
When setting the initial size of a sidesheet, the `isResizing` flag is set to true to allow sidesheets to disable animations for smoother renders, but if the user moves the mouse while setting the initial size the onResize handler will trigger and resize the sidesheet

## Solution
Split `isResizing` into `isResizing` and `isMouseDown`. When the user is resizing, both `isResizing` and `isMouseDown` is set to **true**, but when programatically resizing (initial render with size from local storage) only `isResizing` is set to true

### DIFF
--- a/src/components/general/SideSheet/Standard/styles.less
+++ b/src/components/general/SideSheet/Standard/styles.less
@@ -70,6 +70,7 @@
        display: flex;
        flex-direction: column;
        overflow: auto;
+       flex-grow: 1;
     }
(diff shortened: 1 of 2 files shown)


eskil0312:
Remove console.logs

MartinForre:
@eskil0312 Removed console.log

PR #208, opened 2020-01-03
Slider - hide handle option

eskil0312:


src/components/general/Slider/index.tsx
```diff
@@ -112,7 +117,7 @@ const Slider: React.FC<SliderProps> = ({ value, markers, disabled, onChange }) =
                 }}
             />
             <button
-                className={styles.handle}
+                className={handleClassNames}
```
MartinForre: Wouldn't it be better to not render the handle at all?

eskil0312: Good point, handle will now not render on hide

PR #214, opened 2020-01-09
Try to keep the popover in view

MartinForre:


src/hooks/usePopoverRef/components/Container.tsx
```diff
@@ -52,6 +77,23 @@ const PopoverContainer = React.forwardRef<
...
+
+        const nextPlacement = getAlternativePlacement(
+            preferredPlacement || 'below',
+            testedPlacements
+        );
+
+        setTestedPlacements([...testedPlacements, nextPlacement]);
+        setPlacement(nextPlacement);
+        setIsInView(true);
+    }, [isInView]);
```
sandves: Missing dependency on `testedPlacements`?

MartinForre: No. It should only test a new placement when isInView changes to false, not when testedPlacements changes.

PR #216, opened 2020-01-14
Use enqueueAsyncOperation for continuous breakpoint checking

MartinForre:


src/hooks/useBreakpoint.ts
```diff
@@ -25,19 +25,27 @@ const useBreakpoint = (
     const [currentBreakpoints, setCurrentBreakpoints] = useState<string[]>([]);
     const nodeRef = useRef<HTMLDivElement>(null);
 
-    const performCheckSize = () => {
-        window.requestAnimationFrame(() => {
+    const performCheckSize = async (abortSignal: AbortSignal) => {
+        enqueueAsyncOperation(() => {
             if (!nodeRef.current) {
+                performCheckSize(abortSignal);
```
sandves: Why is this recursive call neseccary?

MartinForre: Good point. The useEffect with dependency on the nodeRef will restart the size checking

PR #217, opened 2020-01-15
Various changes to support new position model

eskil0312:


eskil0312:
Also added custom safe close titles for modal side sheet

PR #222, opened 2020-01-17
Add elipsis to navigation drawer

GryNagel:
This is a suggestion as we have some issues with long text in the navigation drawer:

This is how it looks now when the navigation drawer includes long titles: 
no_ellipsis

And this is how it will look with this PR:
ellipsis


MartinForre:
Perhaps this should be accompanied with a tooltip to display the entire text?

GryNagel:
Yes, a tooltip is a good idea! Should we have tooltips for all items, or only the items with elipsis? I looked at some ways of figuring out if an item has ellipsis, and it should be possible to track, but that means that the tooltip needs to be conditional which is a bit harder as the useTooltipRef is a hook so if you have any suggestions it would be appreciated!

MartinForre:
You could pass an empty string to the useTooltipRef hook. Then it wouldn't render the tooltip at all.

GryNagel:
Figured out a way! Items will now have tooptips if ellipsis is active :)

src/components/general/NavigationDrawer/components/Child.tsx
```diff
@@ -1,10 +1,21 @@
...
+    const textRef = React.useRef<HTMLElement | null>(null);
+    const [shouldHaveTooltip, setShouldHaveTooltip] = useState(false);
+    const tooltipRef = useTooltipRef(title, 'right');
+
+    useEffect(() => {
+        if (textRef.current) {
+            const isOverflowing = textRef.current.offsetWidth < textRef.current.scrollWidth;
+            setShouldHaveTooltip(isOverflowing);
+        }
+    }, [textRef])
```
MartinForre: The dependency should probably be on textRef.current since the ref really doesn't change. Also, what if the title changes? Should there be a dependency on that as well?

GryNagel: When adding textRef.current as a dependency the tooltip does not work. We have observed that with textRef the getNavigationContent is run when you hover the title, but with textRef.current it does not. Not sure why or if this is the reason that the tooltip is only shown with textRef as dependency. Suggestions?

MartinForre: Seems strange 🤔 The ref itself shouldn't change, only the value of the `current` property. Therefore, having a dependency on the ref would only run the effect after initial render. In most cases that's enough since the `current` property is set in between rendering and the effect, but if the component is rerendered with a different DOM node, or the title changes the tooltip would not.

GryNagel: Very strange indeed. Spent some time trying to debug without getting any further.

MartinForre: Allright. As long as it works the way it is now, we'll merge it and figure it out later

GryNagel: I tested some more to make sure, and unfortunately discovered that the section tooltip only works after it's child gets rendered. Group and child tooltips works as they should, but something weird is going on with section :/

MartinForre: @GryNagel My latest commit seems to fix the issue with tooltip on sections

GryNagel: Looks perfect! Thanks (and sorry I completely forgot about this PR)! 👏

PR #225, opened 2020-01-23
Context Selector - Implement new options from the App manifest.

olerichard:
Nullable, if true in manifest. selector will show a X button. This will remove current context.
placeholder, show custom placeholder in popover when selecting context.
Updated story implemeting the new options.
Updated FusionContext with new ContextBaseUrl. Now using CI

### DIFF
--- a/src/components/core/ContextSelector/index.tsx
+++ b/src/components/core/ContextSelector/index.tsx
@@ -85,6 +85,10 @@ const ContextSelector: React.FC = () => {
                 !isOpen && setIsOpen(true);
             }, [isOpen]);
 
+            const placeholder = contextManifest?.placeholder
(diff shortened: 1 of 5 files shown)


src/components/core/ContextSelector/index.tsx
```diff
@@ -125,9 +136,11 @@ const ContextSelector: React.FC = () => {
     const helperText = React.useMemo(
         () =>
             !contexts.length && !isQuerying && !queryText
-                ? 'No contexts, please try the search bar above (Start typing to search)'
+                ? contextManifest?.placeholder
```
MartinForre: Shouldn't the placeholder be a property to the input?

PR #227, opened 2020-01-30
Set max-height on dropdowns based on distance from window top

MartinForre:


src/components/core/Root/styles.less
```diff
@@ -83,10 +83,8 @@
 .overlay {
     position: fixed;
     overflow: visible;
-    width: 0;
-    height: 0;
-    width: 0;
-    height: 0;
+    // width: 0;
+    // height: 0;
```
eskil0312: Remove comment?

PR #231, opened 2020-01-31
Power BI report component

olerichard:
Simple way to implement a report in any app. 

component takes 2 parameters. 
mandatory report ID. ID found in reports app for app you want to show. 
optional report filter. Applies a filter to report. Typically used with context selection.


src/components/data/PowerBIReport/index.tsx
```diff
@@ -0,0 +1,258 @@
...
+    const embed = React.useCallback(
+        (node: HTMLDivElement) => {
+            if (embedInfo) {
+                const config = getConfig(embedInfo.embedConfig.embedType.toLowerCase());
+
+                embeddedRef.current = powerbi.embed(node, config);
+
+                embeddedRef.current.off('loaded');
+                embeddedRef.current.off('error');
+                embeddedRef.current.on('loaded', () => setIsLoading(false));
```
MartinForre: Should we perhapse track time to load metric here?

src/components/data/PowerBIReport/index.tsx
```diff
@@ -150,7 +156,17 @@ const PowerBIReport: React.FC<PowerBIProps> = ({ reportId, filters }) => {
...
+                    // console.log("Loaded Time:",(new Date().getTime() - timeLoadStart.getTime())/1000," Seconds")
+
+                    telemetryLogger.trackMetric({
+                        name: `${useCurrentApp.name}-EmbedloadedTime`,
+                        average: new Date().getTime() - timeLoadStart.getTime(),
+                        sampleCount: 1,
+                    });
+
+                    setIsLoading(false);
+                });
```
olerichard: There is also a "rendered" listener here that could be used for tracking. 
it triggers when the report is done rendering or user has interacted with the report. 

I that case i could add something like this : 

```
embeddedRef.current.on('rendered', () =>{
   telemetryLogger.trackMetric({
     name:`${useCurrentApp.name}-EmbedRenderedTime`,
     average: new Date().getTime() - timeLoadStart.getTime(),
     sampleCount:1
 });
```

src/components/data/PowerBIReport/index.tsx
```diff
@@ -0,0 +1,274 @@
...
+    const [fusionError, setFusionError] = React.useState<FusionError>();
+
+    const [report, setReport] = React.useState<Report>();
+    const [embedInfo, setEmbedInfo] = React.useState<EmbedInfo>();
+    const [accessToken, setAccessToken] = React.useState<AccessToken>();
+    const telemetryLogger = useTelemetryLogger();
+
+    const embedRef = React.useRef<HTMLDivElement>(null);
+    const embeddedRef = React.useRef<pbi.Embed | null>(null);
+    const timeLoadStart = new Date();
```
MartinForre: This should probably be stored with useRef or useState. Just setting it to a variable will reset it each render

PR #266, opened 2020-02-10
Reapply filter when clicking on "reset filter"

olerichard:
When user clicks on the "reset filter" button all filters are removed from the report. 
Including selected context.  Removing context should not be possible. 

A onClickButtonClick event handler has been added to handle this problem.
If reset button is clicked filter will be reapplied to report.

### DIFF
--- a/src/components/data/PowerBIReport/index.tsx
+++ b/src/components/data/PowerBIReport/index.tsx
@@ -23,6 +23,7 @@ import {
 } from './models/ReportLevelFilters';
 
 import * as styles from './styles.less';
+import { ButtonClickEvent } from './models/EventHandlerTypes';
 
 type PowerBIProps = {
     reportId: string;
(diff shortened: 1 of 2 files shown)


src/components/data/PowerBIReport/index.tsx
```diff
@@ -198,9 +199,13 @@ const PowerBIReport: React.FC<PowerBIProps> = ({ reportId, filters }) => {
     React.useEffect(() => {
         if (embeddedRef.current) {
             embeddedRef.current.off('pageChanged');
+            embeddedRef.current.off('buttonClicked');
             embeddedRef.current.on('pageChanged', () => {
                 setFilter();
             });
+            embeddedRef.current.on('buttonClicked', (button: ICustomEvent<ButtonClickEvent>) => {
+                if (button.detail.title === 'reset filter') setFilter();
```
MartinForre: Should we ensure casing here?

PR #287, opened 2020-02-12
Support for showing assignee's for rotation positions

eskil0312:
Org chart and position card

### DIFF
--- a/src/components/org/OrgChart/components/Aside.tsx
+++ b/src/components/org/OrgChart/components/Aside.tsx
@@ -15,6 +15,7 @@ function Aside<T>() {
             cardWidth,
             numberOfCardsPerRow,
             width,
+            additionalAsideRowHeight,
         },
         dispatch,
     } = useContext<OrgChartContextReducer<T>>(OrgChartContext);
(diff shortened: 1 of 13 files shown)


src/components/org/OrgChart/components/Aside.tsx
```diff
@@ -62,24 +63,68 @@ function Aside<T>() {
...
+            }
+            return (
+                Math.ceil(greatestAdditionPersons / 3) *
+                (numberOfCardsPerRow === 1 ? rowMargin - 20 : rowMargin)
+            );
+        },
+        [numberOfCardsPerRow, rowMargin]
+    );
+
+    const additionalRowHeight = rows.reduce(
```
MartinForre: This should probably be memoized?

src/components/org/OrgChart/components/Children.tsx
```diff
@@ -68,27 +72,75 @@ function Children<T>() {
...
+            }
+            return (
+                Math.ceil(greatestAdditionPersons / 3) *
+                (numberOfCardsPerRow === 1 ? rowMargin - 20 : rowMargin)
+            );
+        },
+        [numberOfCardsPerRow, rowMargin]
+    );
+
+    const additionalRowHeight = rows.reduce(
```
MartinForre: Memoize this as well?

src/components/org/OrgChart/components/Aside.tsx
eskil0312: Memorized and moved common logic to a hook

PR #325, opened 2020-02-18
Adaptive card viewer

eskil0312:
React component wrapper for adaptive cards

### DIFF
--- a/src/components/general/AdaptiveCardViewer/index.tsx
+++ b/src/components/general/AdaptiveCardViewer/index.tsx
@@ -32,6 +32,7 @@ const AdaptiveCardViewer: React.FC<AdaptiveCardViewerProps> = ({
     onActionSubmit,
     className,
 }) => {
+    const cardContainerRef = React.useRef<HTMLDivElement | null>(null);
     const adaptiveCard = React.useMemo(() => new AdaptiveCards.AdaptiveCard(), []);
 
     AdaptiveCards.AdaptiveCard.onProcessMarkdown = (text, result) => {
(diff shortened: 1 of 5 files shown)


src/components/general/AdaptiveCardViewer/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+    const appendNewChild = React.useCallback(
+        (n: HTMLElement) => {
+            n.firstChild && n.removeChild(n.firstChild);
+            n.appendChild(result);
+        },
+        [result]
+    );
+
+    const jsxAdaptiveCard = React.useMemo(
+        () => <div className={className} ref={n => n && appendNewChild(n)} />,
```
MartinForre: Is this the recommended way of rendering adaptive cards in react? Would think that using useEffect to append childs to the div would be a more "react-way" of doing it?

eskil0312: I didn't find any documentation for rendering adaptive cards in react, just in JS. 
Now uses useEffect to append child to the container div

src/components/general/AdaptiveCardViewer/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+        adaptiveCard.parse(payload);
+        return adaptiveCard.render();
+    }, [payload]);
+
+
+    React.useEffect(() => {
+        if (cardContainerRef && cardContainerRef.current) {
+            const current = cardContainerRef.current;
+            current.firstChild && current.removeChild(current.firstChild);
+            current.appendChild(result);
```
MartinForre: Instead of removing the child _before_ appending the new one, we should remove the child in the cleanup method for the effect. That would make the entire process "async" instead of blocking

eskil0312: Now removes child in useEffect cleanup function

PR #339, opened 2020-02-24
Added default styling to adaptive card

eskil0312:


src/components/general/AdaptiveCardViewer/defaultHostConfig.ts
```diff
@@ -0,0 +1,242 @@
+import { styling } from '@equinor/fusion-components';
+const gridUnit = (unit: number) => {
```
MartinForre: This should probably be added to the styling module?

PR #340, opened 2020-02-24
Set/exchange context when switching app

MartinForre:


src/components/core/ContextSelector/index.tsx
```diff
@@ -121,6 +122,36 @@ const ContextSelector: React.FC = () => {
...
+        setIsOpen(true);
+    }, [contextManager, currentContext]);
+
+    React.useEffect(() => {
+        if(!contextManifest || !currentContext) {
+            return;
+        }
+
+        if(contextManifest.types.indexOf(currentContext.type.id)) {
+            contextManager.setCurrentContextAsync(currentContext);
```
eskil0312: This will also trigger if `contextManifest.types.indexOf(currentContext.type.id) === -1`, is that correct behavior?

MartinForre: Good catch :+1:

PR #361, opened 2020-02-27
Added if initialPerson is set to null, then clear section

thomasFjorstad:
If the prop `initialPerson` in PersonPicker is set, it is not possible to clear it out.
This is a fix for it.


MartinForre:
What's the motivation behind this change?

thomasFjorstad:
If the prop initialPerson in PersonPicker is set, it is not possible to clear it out.
This is a fix for it.

PR #377, opened 2020-03-11
Add Chip component

hgut-eq:


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@equinor/fusion-components",
-    "version": "1.1.53",
+    "version": "1.1.54",
```
MartinForre: Version of npm packages should not be bumped in a PR.
1. Merge the PR to master
2. Pull master
3. npm version patch
4. git push --follow-tags
5. Create release from version tag in github

hgut-eq: Woops, OK! Reverted the version now

MartinForre: There's no changes to the version property in package.json now, so it should be fine to merge

PR #429, opened 2020-03-26
simple write out contexttype in input field

olerichard:


src/components/core/ContextSelector/index.tsx
```diff
@@ -61,16 +67,16 @@ const ContextSelector: React.FC = () => {
...
 
             const selectedValue = React.useMemo(() => {
                 if (isOpen) {
                     return queryText;
                 } else if (selectedItem) {
-                    return selectedItem.title;
+                    return `${selectedItem.title} (${selectedItem.contextType.id})`;
                 } else if (currentContext) {
-                    return currentContext.title;
+                    return `${currentContext.title} (${currentContext.type.id})`;
```
MartinForre: Should we use the function you made to map context type to "human readable"? E.g. `ProjectMaster `-> `Project master`

PR #438, opened 2020-03-27
Add disabled prop and styling to datepicker

GryNagel:
- Add prop
- Make calendar icon grey when disabled
- Check for disabled in click function


eskil0312:
It looks good, but some of these diffs are a bit weird. Lines of code that are just switches place 😄  Do you have any idea why these diffs occur? 🤔

GryNagel:
The props gets sorted, that is why the diff looks a bit of. The sorting addon is turned off by default, but it still ran on save, and I could not figure out why 😅

PR #451, opened 2020-03-31
Updated account type tool tip

eskil0312:


HansDahle:
Did we not talk about including the more descriptive text as well in the tool tip?

eskil0312:
@HansDahle  I don't recall talking about it, but I guess that could be added as well

HansDahle:
> 
> 
> @HansDahle I don't recall talking about it, but I guess that could be added as well

Should have read the userstories ;)

eskil0312:
> > @HansDahle I don't recall talking about it, but I guess that could be added as well
> 
> Should have read the userstories ;)

😄

PR #452, opened 2020-03-31
searchable drop down hide textoverflow when using components

olerichard:
This change is applied directly to the SearchableDropdown component.
and in that regard have abit of a far reaching effect, but the concequenses should be minor. 

I have added a text-overflow propery to the searchable dropdown. 
This only takes effect when you pass along a "selectedComponent". 
The overflow will cut the componenet if its to long according to the input field length. 

This fix also takes the dropDownArrow into account. 
It should not be covered by the component if its to long. 
Nor will it resize if the components starts taking up to much space. 
Arrow should be locked to its normal 24x24 size regardless of selected item size.

This change is consistent with the behaviour of the DropDown when not rendering a component inside. 

This issue was found in the resource app, see bug 10755


src/components/general/SearchableDropdown/styles.less
```diff
@@ -8,6 +8,9 @@
 .selectedItem {
     flex-grow: 2;
     padding-left: calc(var(--grid-unit) * 2px);
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: clip;
```
MartinForre: We usually use text-overflow: ellipsis;

PR #504, opened 2020-04-21
Ranged slider

eskil0312:


src/index.ts
```diff
@@ -63,6 +63,7 @@ export { default as Calendar } from './components/general/Calendar';
 export { default as Chip } from './components/general/Chip';
 export { default as DatePicker } from './components/general/DatePicker';
 export { default as Slider } from './components/general/Slider';
+export { default as RangedSlider } from './components/general/Slider/RangeSlider';
```
MartinForre, line 66: While we're here: Should we export the SliderMarker type aswell?

PR #506, opened 2020-04-21
Ranged slider - calculation bugfix

eskil0312:


src/components/general/Slider/RangeSlider.tsx
```diff
@@ -120,7 +116,12 @@ const Slider: React.FC<SliderProps> = ({ values, markers, disabled, hideHandle,
                 className={classNames(styles.slider, styles.active)}
                 style={{
                     left: calculatePosition(lastValue > firstValue ? firstValue : lastValue),
-                    width: diffWidth,
+                    right: `${
+                        100 -
+                        +calculatePosition(lastValue > firstValue ? lastValue : firstValue).split(
```
MartinForre, line 121: Would be better and more readable to use `parseInt(calculatePosition(...), 10)` instead of `+calculatePosition(...).split('%')[0]` i guess. Either that or change calculatePosition to return a number and append the '%' where needet (e.g. line 118)

eskil0312: Did the first one, if i did the second one I've have to append % allot of places

PR #552, opened 2020-05-18
Poc - Person presence indicator

eskil0312:


src/components/org/OrgChart/OrgChart.stories.tsx
```diff
@@ -154,7 +154,42 @@ const position: Position = {
                     name: 'Equinor',
                 },
                 contracts: [],
-                positions: [],
+                positions: [
```
eskil0312, line 157: Positions property was added for testing right? If that's the case, we don't need it

ingridklepsvik: Removed it

src/components/people/PersonDetail/index.tsx
```diff
@@ -1,54 +1,132 @@
...
+
+        try {
+            const response = await apiClients.people.getPresenceAsync(mail, '1.0-preview');
+            setPresence(response.data);
+        } catch (e) {
+            setPresenceError(e);
+        } finally {
+            setIsFetchingPresence(false);
+        }
+    }, []);
```
eskil0312, line 46: Maybe add apiClients to dependency array?

src/components/people/PersonPhoto/__stories__/PersonPhoto.stories.tsx
```diff
@@ -73,6 +73,14 @@ const sizes = {
     small: 'small',
 };
 
+/* const presenceStatus = {
+    available: 'available',
+    away: 'away',
+    doNotDisturb: 'doNotDisturb',
```
eskil0312, line 79: Remove comment?

src/components/people/PersonDetail/index.tsx
ingridklepsvik: Added apiClients to dependency array

src/components/people/PersonPhoto/__stories__/PersonPhoto.stories.tsx
ingridklepsvik: Removed comment

PR #560, opened 2020-05-22
Filter Pane - custom component in header 

olerichard:
FilterPane header is now sticked to the top when scrolling.
FilterPane Header can now receive a component.
It will be aligned on the opposite side of the collapse-arrow.
Component will be hidden when collapsing filterpane.

Added a small filter header component to the Filter Pane story. Demoing how this works.
It just flips the filter pane placement back and forth from right to left.

### DIFF
--- a/src/components/general/FilterPane/index.tsx
+++ b/src/components/general/FilterPane/index.tsx
@@ -37,6 +37,7 @@ export type FilterPaneProps<T> = {
     onChange: OnFilterChangeHandler<T>;
     screenPlacement?: 'right' | 'left';
     onToggleCollapse?: (isCollapsed: boolean) => void;
+    headerComponent?: React.ReactNode;
 };
 
 function FilterPane<T>({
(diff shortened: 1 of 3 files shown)


src/components/general/FilterPane/index.tsx
```diff
@@ -106,9 +108,10 @@ function FilterPane<T>({
                     <div className={styles.collapseExpandButtonContainer}>
                         <CollapseExpandButton onClick={toggleCollapsed} />
                     </div>
+                    {!isCollapsed && headerComponent}
```
MartinForre, line 111: This should perhaps be wrapped in a div with `flex-grow: 1;` or something and ensure that it's contained within the header

olerichard: I have set the header to "push" the collapseButton and headercomponent appart. 
setting headerComponet to grow, breaks this abit. 
Would need some changes to justify the content internally then and checking if its left or right, to justify it to start / end. 
But I dont think it does much either way when testing it. 

As it stands now. 
If the component is to wide. It will just disappear outside, as overflow is hidden. 
If its to tall it will grow the header in height.

PR #629, opened 2020-06-29
Quick Facts

MartinForre:


src/customElements/applicationGuidance/Anchor/index.ts
```diff
@@ -0,0 +1,185 @@
...
+                window.postMessage(message, window.location.origin);
+            }
+
+            this.animationFrame = window.requestAnimationFrame(this.calculateRects);
+        });
+    };
+
+    render() {
+        return html` <slot></slot> `;
+    }
```
odinr, line 131: this element does not seem to need any shadowroot
`createRenderRoot() {
  return this;
}`

src/customElements/applicationGuidance/Wrapper/editQuickFact.ts
```diff
@@ -0,0 +1,170 @@
...
+        );
+    };
+
+    renderActions() {
+        return html`
+            <div class="toolbar-header">
+                <button
+                    class="button border"
+                    ?disabled="${this.isSaving}"
+                    @click="${this.handleCancel}"
```
odinr, line 126: *opt*
by using the decorator `@eventOptions({passive:true}:  AddEventListenerOptions)` above the handler you can control the event args.
Also to make a custom event pass through shadow DOM boundaries, you must set both the `composed` and `bubbles` flags to `true`:

src/customElements/applicationGuidance/Wrapper/index.ts
```diff
@@ -0,0 +1,351 @@
...
+            this.fetchUpdatedQuickFact(this.getScopePath(), this.activeAnchorId);
+        }
+
+        if (!this.isActive && changedProperties.has('isActive')) {
+            this.scopes = [];
+        }
+    }
+
+    disconnectedCallback() {
+        window.removeEventListener('message', this.handleMessage);
```
odinr, line 108: **Other CustomElements that subscribes to message should also implement this.**
*note*
subscription is attached in constructor and not connectedCallback, this might cause issues when react updates virtual dom (not listening for messages after reinserted).

src/customElements/applicationGuidance/Wrapper/interactiveAnchor.ts
```diff
@@ -0,0 +1,84 @@
...
+        `;
+    }
+}
+
+type ApplicationGuidanceInteractiveAnchorHTMLAttributes = ApplicationGuidanceInteractiveAnchorProps &
+    React.HTMLAttributes<ApplicationGuidanceInteractiveAnchor>;
+
+declare global {
+    interface HTMLElementTagNameMap {
+        'fusion-person-photo': ApplicationGuidanceInteractiveAnchor;
```
odinr, line 67: :)

odinr:
- All custom elements should use the same prefix 'fusion', example 'fusion-app-guide-anchor'
- Elements that does not require markup shouldn't have shadow dom
- Subscriptions should attach and dispose on lifecycle callbacks (and be passive)

PR #711, opened 2020-08-13
PowerBI Report RLS update.

olerichard:
- Update fusion api package
  - updated report component to reflect changes in fusion api package.
  - api update mostly affects model changes.
  - Uses of any has been removed.
  - new RLS Error handling
  - RLS will not be accounted for and will show its own error message when user dont have access.
  - Other errors still uses old standard error message.
  - loading spinner implementations has been reworked.
  - the clause for showing spinner didnt acutally make much sense and has been corrected.
  - also added a chaning loading message. Fetching toking can sometimes take abit of time. This call also retries on fail. This information atleast gives the user a sense of whats going on. Also its usefull information for us if its a problem that gets reported in.


src/components/data/PowerBIReport/components/RlsErrorMessage/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+        const fetchedDescriptions = await reportApiClient.getDescription(report.id);
+
+        setNoAccessMessage(fetchedNoAccessMessage?.data || null);
+        setDescription(fetchedDescriptions?.data || null);
+        setRequirements(fetchedRequirements?.data || null);
+        setIsFetching(false);
+    };
+
+    if (isFetching)
+        return <Spinner title={'Retrieving report access specifications '} floating centered />;
```
odinr, line 50: the effect is only triggered on initial load, changing report will not fetch associated requirements.
there is no try/catch/final on awaited promises, so any api errors will always show spinner.

src/components/data/PowerBIReport/components/RlsErrorMessage/index.tsx
```diff
@@ -32,18 +32,22 @@ const RlsErrorMessage: React.FC<RlsErrorMessageProps> = ({ report }) => {
...
-        setNoAccessMessage(fetchedNoAccessMessage?.data || null);
-        setDescription(fetchedDescriptions?.data || null);
-        setRequirements(fetchedRequirements?.data || null);
-        setIsFetching(false);
+            setNoAccessMessage(fetchedNoAccessMessage?.data || null);
+            setDescription(fetchedDescriptions?.data || null);
+            setRequirements(fetchedRequirements?.data || null);
+            setIsFetching(false);
+        } catch {
+            setIsFetching(false);
```
odinr, line 49: `try{} catch{} finally{setIsFetching(false);}`

PR #807, opened 2020-09-23
feat: add disabled state to navigation drawer

eskil0312:


odinr (approved):
Use the imperative, present tense: "add" not "added" nor "adds" in messages :)

PR #820, opened 2020-10-05
fix: Date picker update state on blur and tab press

eskil0312:
Also changed default image size for adaptive cards

### DIFF
--- a/src/components/general/DatePicker/index.tsx
+++ b/src/components/general/DatePicker/index.tsx
@@ -69,7 +69,7 @@ const DatePicker: React.FC<DatePickerProps> = ({
                     setIsOpen(false);
                 }
             },
-            [isOpen, tryParseDate, overlayContainer]
+            [isOpen, tryParseDate, overlayContainer, isValidMask]
         );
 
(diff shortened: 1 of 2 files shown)


src/components/general/DatePicker/index.tsx
```diff
@@ -33,25 +40,54 @@ const DatePicker: React.FC<DatePickerProps> = ({ disabled, error, errorMessage,
...
-                    const parsedDate = parseDate(maskedValue);
-                    onChange(parsedDate);
+        const tryParseDate = useCallback(() => {
+            try {
+                const parsedDate = parseDate(maskedValue);
+                onChange(parsedDate);
+            } catch (e) {}
+        }, [maskedValue, parseDate]);
+
+        const handleKeyUp = useCallback(
```
odinr, line 50: when only using codes, please provide in comment which key has been use.
tip: import lib for mapping name => keycode

odinr (approved):

src/components/general/DatePicker/index.tsx
eskil0312: Replace deprecated `KeyboardEvent.keyCode` with `KeyboardEvent.key`

PR #840, opened 2020-10-27
Notification side sheet redesign

ingridklepsvik:
Added tabs to notification side sheet. 

feat: changed notification side sheet

### DIFF
--- a/src/components/general/NotificationsSideSheet/index.tsx
+++ b/src/components/general/NotificationsSideSheet/index.tsx
@@ -1,10 +1,12 @@
 import * as React from 'react';
 import * as styles from './styles.less';
-import { ModalSideSheet, Chip, Button, Spinner, styling } from '@equinor/fusion-components';
+import { ModalSideSheet, Chip, Button, Spinner, styling, Tab, Tabs, SettingsIcon } from '@equinor/fusion-components';
 import { NotificationCard, useGlobalNotificationCardsActions } from '@equinor/fusion';
 import NotificationDateDivisions from './NotificationDateDivisions';
 
 type NotificationsSideSheetProps = {
+    settings: React.ReactElement | null;
+    showSettings?: boolean;
     open: boolean;
     onClose: () => void;
     notifications: NotificationCard[];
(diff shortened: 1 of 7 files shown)


src/components/general/NotificationsSideSheet/index.tsx
```diff
@@ -13,59 +15,31 @@ type NotificationsSideSheetProps = {
...
 type NotificationsSideSheetContentProps = {
+    settings: React.ReactElement | null;
+    showSettings?: boolean;
     readNotifications: NotificationCard[];
     unReadNotifications: NotificationCard[];
     isFetchingReadNotifications: boolean;
     isFetchingUnReadNotifications: boolean;
 };
 
 const NotificationsSideSheetContent: React.FC<NotificationsSideSheetContentProps> = ({
```
odinr, line 26: this file contains two components, there should only be one/file, split into 2 files

src/components/general/NotificationsSideSheet/index.tsx
```diff
@@ -78,6 +52,70 @@ const NotificationsSideSheet: React.FC<NotificationsSideSheetProps> = ({
...
+                                        <Button
+                                            onClick={onReadAll}
+                                            disabled={unReadNotifications.length <= 0}
+                                        >
+                                            {isMarkingNotifications ? <Spinner inline /> : 'Mark all as read'}
+                                        </Button>
+                                    <NotificationDateDivisions notifications={unReadNotifications} />
+                                    </>
+                                ))}
+                    </div>
```
odinr, line 84: too much logic inside render method, either refactor to callback/memo or create a new component(s)

odinr (changes requested):

src/components/general/NotificationsSideSheet/index.tsx
ingridklepsvik: split the two components into two files

src/components/general/NotificationsSideSheet/index.tsx
ingridklepsvik: Removed some logic from the render method and refactored to memo

odinr (approved):

PR #861, opened 2020-11-10
Power BI Report  - Reembed report fix 

olerichard:
New context access check revield an issue when reembedding a report.
If a report was embed. Then the embed was removed and later reembedded.
The embed would fail.
This has been solved by splitting up the component, ensuring the embed gets a full reset when it get embedded.

The entry point will now only handle fetching of report data and access checks and error handling. Everything related to Fusion.
If user has access and all data is found.
The component will use the new ReportEmbed component.
This handles everything PowerBI.
It also implement the bookmarks manager.

The code and implementation should be mostly the same here.
Its only the big index file that has been split up into two.
Making this component abit easier to reason about, as the initial compoent had grown quite large.


olerichard:
redundant

PR #863, opened 2020-11-13
fix: apply awaitable bookmark 

eskil0312:
Making sure the pageChange callback, triggering the setFilter, is updated by the awaitable bookmark.
Also re-added the check for fusion error or report error. Without these checks I got no access even trough I had access

### DIFF
--- a/src/components/data/PowerBIReport/index.tsx
+++ b/src/components/data/PowerBIReport/index.tsx
@@ -336,12 +336,8 @@ const PowerBIReport: React.FC<PowerBIProps> = ({ reportId, filters, hasContext }
             }
         }
     }, [embedRef, accessToken, isFetching]);
-
-    if (
-        (powerBIError && powerBIError.detail.errorCode) ||
-        (fusionError && fusionError.fusionError?.error?.code)
-    ) {
-        //Only handling selected errors from Power BI. As you migh get errors that can be ignored.
+    if (powerBIError || fusionError?.fusionError) {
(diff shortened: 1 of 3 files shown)


odinr (approved):

src/components/data/PowerBIReport/index.tsx
```diff
@@ -336,9 +336,8 @@ const PowerBIReport: React.FC<PowerBIProps> = ({ reportId, filters, hasContext }
             }
         }
     }, [embedRef, accessToken, isFetching]);
-
-    if (powerBIError || fusionError) {
-        //Only handling selected errors from Power BI. As you migh get errors that can be ignored.
+    if (powerBIError || fusionError?.fusionError) {
```
olerichard, line 339: Are you sure about this ? 
There is a new access check here. If fusionError is set, the error should be valid.

olerichard (changes requested):

src/components/data/PowerBIReport/index.tsx
olerichard: see checkContextAccess. 
If you are testing with a report that has RLS config, this runs.
If you dont have access to the external ID, you might end up with no access.

eskil0312: For me, that is not the case. When I tried to access the analytics reports the `fusionError` was set but `fusionError.fusionError` and `fusionError.status` was undefined

olerichard (approved):

PR #893, opened 2020-12-15
Garden - update columns fix

eskil0312:
Seems like the local cache is not updating on item change, added proposed solution.
NOT TESTED YET


olerichard:
renderItem and item is not the same type off object.  Its very unlikely they have any keys in common. 
So this wont really do anything. 

There is a way to fix this in the implementation of the component in the apps. 
And therefore ill close this PR.

PR #896, opened 2020-12-16
Fix/presence teams

ingridklepsvik:


src/components/people/PersonPhoto/PersonPresenceIcon/index.tsx
```diff
@@ -8,32 +8,40 @@ import {
...
         styles.presenceContainer,
         useComponentDisplayClassNames(styles),
         styles[size],
-        styles[presence.toLowerCase()]
+        styles[presenceAvailability?.toLowerCase()]
     );
 
     const presenceIcon = React.useMemo(() => {
-        if (presence === 'Online') {
+        if (presenceAvailability === 'Available' || 'AvailableIdle') {
```
olerichard, line 29: This probably dosnt work as it should, same issue on line 35
Needs to be similar to how it is on line 32

olerichard (changes requested):

olerichard (approved):

odinr (approved):

PR #969, opened 2021-03-09
Feature/export models

tobiasny:


src/index.ts
```diff
@@ -80,6 +80,13 @@ export { default as Chip } from './components/general/Chip';
 export { default as DatePicker } from './components/general/DatePicker';
 export { Slider, RangedSlider, SliderMarker } from './components/general/Slider';
 export { default as RequestWorkflow } from './components/general/RequestWorkflow';
+export {
+    Workflow,
+    WorkflowStep,
+    ProvisioningStatus,
+    ProvisioningState,
+    Person
```
eskil0312, line 88: I think we need to be more specific with the naming when we are exporting these models. Person could be WorkflowPerson or RequestWorkflowPerson, same with Provisioning

PR #1033, opened 2021-04-22
Feature/general/bookmarksmanager

jaysencpp:


jaysencpp:
odata-query from fusion-api causes this build to fail (but doesn't fail when i build it locally). I guess because we deleted the `declare module odata-query` from fusion-api?

odinr (approved):

PR #1079, opened 2021-05-12
Bookmarks manager feedback changes

jaysencpp:
* When entering a shared bookmark URL (.../apps/....?bookmarkId=xxxx), the payload (e.g. a filter) is automatically applied to the data showed in the app.
* Changed action button label for one modal to a better label. 
* When unsharing a bookmark, a modal will appear asking if you are sure you want to unshare the bookmark. 
* Fix Filter issue where checkboxes and radio buttons wont update when updating filter from bookmark.

### DIFF
--- a/src/components/general/BookmarksManager/BookmarkSideSheet/SideSheetManager/Bookmark/index.tsx
+++ b/src/components/general/BookmarksManager/BookmarkSideSheet/SideSheetManager/Bookmark/index.tsx
@@ -46,15 +46,11 @@ export const Bookmark = ({
         });
         if (!response.confirmed) return;
 
-        try {
-            store.deleteBookmark(bookmark.appKey, bookmark.id);
-        } catch (e) {}
+        store.deleteBookmark(bookmark.appKey, bookmark.id);
     };
 
     const handleRemove = () => {
-        try {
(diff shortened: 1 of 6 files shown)


src/components/general/BookmarksManager/BookmarkSideSheet/SideSheetManager/Bookmark/index.tsx
```diff
@@ -110,6 +104,27 @@ export const Bookmark = ({
                 hideCancelAction: true,
                 body: <ShareBody />,
             });
+            try {
+                store.updateBookmark(bookmark.id, {
```
odinr, line 108: havent looked in store, but does this function throw error?

odinr (changes requested):

src/components/general/BookmarksManager/BookmarkSideSheet/SideSheetManager/Bookmark/index.tsx
jaysencpp: In the epic for updateBookmark there is `catchError((error) => of(failure({ action, error }))),` - so maybe no need for the try catch since it's already handled in the epic?

odinr (approved):

PR #1088, opened 2021-05-19
Feat/person picker - custom section generator

tobiasny:


src/components/people/PersonPicker/index.tsx
```diff
@@ -25,6 +25,12 @@ type PersonPickerProps = {
     hasError?: boolean;
     errorMessage?: string;
     onSelect?: (person: PersonDetails) => void;
+    customSectionGenerator?: (
```
odinr, line 28: make a type or interface for PersonPickerSectionFn
in props: {sectionFn = peopleToSection}
in effect only call sectionFn

odinr (approved):

PR #1093, opened 2021-05-26
check for contextless bookmarks

jaysencpp:
A fix for the bookmarks managers in apps where there is no context. The bookmarks will be put under an accordion called "Global" if there is no context.

### DIFF
--- a/src/components/general/BookmarksManager/BookmarkSideSheet/SideSheetManager/index.tsx
+++ b/src/components/general/BookmarksManager/BookmarkSideSheet/SideSheetManager/index.tsx
@@ -42,7 +42,7 @@ const createGroupByContextId = (allBookmarks: BookmarkListResponse[]) => {
     allBookmarks.reduce((__prev, curr) => {
         const key = curr.context ? curr.context.id : 'Global';
         if (!groupedContexts[key]) {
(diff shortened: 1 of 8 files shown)


src/components/general/BookmarksManager/BookmarkSideSheet/SideSheetManager/index.tsx
```diff
@@ -50,7 +52,7 @@ const createGroupByContextId = (allBookmarks: BookmarkListResponse[]) => {
 const createGroupByContext = (allBookmarks: BookmarkListResponse[]) => {
     const groupedContext: Record<string, BookmarkListResponse[]> = {};
     allBookmarks.reduce((__prev, curr) => {
-        const key = curr.context ? curr.context.id : 'unknown';
+        const key = curr.context ? curr.context.id : 'Global';
```
odinr, line 55: when checking, use either constant or enum

odinr (changes requested):

src/components/general/BookmarksManager/BookmarkSideSheet/SideSheetManager/index.tsx
jaysencpp: extracted the string into a const

odinr (approved):

PR #1103, opened 2021-05-31
Fix: Filter pane section focus issue

tobiasny:


src/components/general/FilterPane/components/Section.tsx
```diff
@@ -38,19 +38,18 @@ function Section<T>({ terms, filterCount, section, onChange, quickFactScope }: S
...
     );
 
-    const renderedFilterComponents = section.filters.map((filter) => {
-        const term = terms.find((term) => term.key === filter.key);
-        return (
+    const renderedFilterComponents = useMemo(() => {
+        return section.filters.map((filter) => (
             <Filter
-                key={`${filter.key}_${term?.value}`}
+                key={filter.key}
```
jaysencpp, line 44: I changed this key in a previous PR because `key={filter.key}` will cause a key warning, and the checkboxes and radio buttons in the filter pane won't update when e.g. applying a bookmark filter.

odinr (approved):

PR #1111, opened 2021-06-03
Feat - Workflow Component - Display a person in a pending workflow step

tobiasny:
Includes an option to display a person in a specified workflow step with pending state.

E.g display the resource owner of a proposed person in the proposal step if it is still pending

### DIFF
--- a/src/components/general/RequestWorkflow/models.ts
+++ b/src/components/general/RequestWorkflow/models.ts
@@ -1,6 +1,8 @@
-import { PersonAccountType } from "@equinor/fusion";
+import { PersonAccountType } from '@equinor/fusion';
 
 export type WorkflowProvisioningState = 'NotProvisioned' | 'Provisioned' | 'Error' | 'Unknown';
+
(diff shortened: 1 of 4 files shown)


src/components/general/RequestWorkflow/models.ts
```diff
@@ -17,17 +19,22 @@ export type WorkflowPerson = {
...
 
 export type WorkflowStep = {
     id: string;
-    state: 'Pending' | 'Active' | 'Approved' | 'Skipped';
+    state: WorkflowStepState;
     name: string;
     description: string;
-    started: Date | null; 
-    completed: Date | null; 
+    started: Date | null;
```
eskil0312, line 34: How about we populate the WorkflowStep with a pendingPerson parameter instead of having pendingPerson in a separate prop? We will then have to manipulate in workflow object that we get from the api in the app. Something like
`workflow.steps.map(s => s.id === "proposal" ? {...s, pendingPerson: "azureUniqueID"} : s)`

tobiasny: Good idea!

eskil0312 (approved):

PR #1119, opened 2021-06-04
Fix: PersonPhoto hook issue

tobiasny:


odinr (approved):

eskil0312:
Can you do the same fix in the PersonCard and PersonDetails component?

PR #1135, opened 2021-06-14
Fix context selector input

odinr:
when selecting text of chosen context, the input field is not shown


.storybook/authTokenAddon.jsx
```diff
@@ -25,8 +25,8 @@ const AuthTokenInput = ({ active }) => {
                 margin: 24,
             }}
         >
-            <button onClick={login}>Sign in</button>
-            <p>Or paste auth token (without bearer)</p>
+            {/* <button onClick={login}>Sign in</button> */}
```
eskil0312, line 28: Remove comment?

src/components/core/ContextSelector/index.tsx
```diff
@@ -104,34 +104,52 @@ const ContextSelector: FC = () => {
...
                 return contextManifest?.placeholder
                     ? contextManifest.placeholder
                     : 'Search context';
             }, [contextManifest?.placeholder]);
 
+            const hasFocus = inputRef.current === document.activeElement;
+
             return (
-                <>
+                <div style={{ display: 'flex', alignItems: 'center' }}>
```
eskil0312, line 119: Avoid inline styling

src/components/core/ContextSelector/index.tsx
```diff
@@ -104,34 +104,52 @@ const ContextSelector: FC = () => {
...
+                            onKeyUp={onKeyUpCloseDropDown}
+                            placeholder={placeholder}
+                            style={{ opacity: hasFocus ? 1 : 0 }}
+                            className={styles.searchContainer}
+                            ref={inputRef}
+                        />
+                        <span
+                            className={classNames(styles.searchInput, styles.overlay)}
+                            style={{
+                                opacity: hasFocus ? 0 : 1,
```
eskil0312, line 136: Avoid inline styling

odinr: inline styling is not bad, this is better than adding 2 classnames for just toggling opacity.
when react parse the style from the virtual dom to the browser dom, this inline style would be shorter than the hashed generated classname for this element. Also the browser need to do a lookup for the marked class of the element and apply style, consider all the garbage style generated inside head, this is a costly operation. 

In the situation where you have multiple elements, like td inside a table, the parse size would most likely be greater and therefor have a longer render time on update.

normally one would just select the element and toggle the css property, but in React, using a Ref is also very costly.

Using classes is a good rule of thumb, especially considering reuse on multiple elements to optimize rendering, but there is also exception to this rule

eskil0312 (approved):

PR #1147, opened 2021-06-21
fix: usePopoverRef - stop setIsPopoverOpen when using hover

olerichard:
useHoverToogleController does not return a setState handler. Causing useEffect to fail when, using the hook.

Added a handler where, it just returns if its a hover, and a redundant check on weither setIsPopoverOpen when calling function.

### DIFF
--- a/src/hooks/usePopoverRef/index.tsx
+++ b/src/hooks/usePopoverRef/index.tsx
@@ -37,7 +37,9 @@ export default <T extends HTMLElement>(
         : useClickToggleController<T>();
 
     useEffect(() => {
-        setIsPopoverOpen(isOpen);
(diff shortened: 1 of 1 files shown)


src/hooks/usePopoverRef/index.tsx
```diff
@@ -37,7 +37,9 @@ export default <T extends HTMLElement>(
         : useClickToggleController<T>();
```
olerichard, line 37: setIsPopoverOpen only exsist on the useClickToggleController. Caused the useEffect to crash when hover was true

olerichard:
#1146 usePopoverRef setState bug fixes this bug. 

Closing this PR

PR #1174, opened 2021-07-19
feat: 30% opacity when inactive prop PersonDetails

sstangeland:
https://dev.azure.com/statoil-proview/Fusion/_backlogs/backlog/Fusion%20Team/Epics/?showParents=true&workitem=20992

As linked task implies, there is a need to indicate if a persons account is inactive.


sstangeland:
Backend work is needed to provide a inactive flag (or similar) on and prop set on PersonDetails in fusion-api

Issue #1165, opened 2021-06-30
[DropDownController] - Does not close when triggering another controller


odinr:
when showing dropdown content and trigger another dropdown the scrim in not triggered

PR #1183, opened 2021-08-11
fix: dropdown and scrim refs and click handlers

### LINKED ISSUE #1166: [SideSheet] - Scrim click
when pressing the mouse inside the modal and release the button on the scrim, the modal will trigger scrim click

fix: add flag for scrim click on mouse down

jaysencpp:
Fixes #1166 and #1165 

#1165 issue was because the side sheet component had a `stopPropagation()` and the click handler from `useClickOutsideOverlayPortal` wouldn't be added correctly. 
The logic inside Dropdown component now is to add an event listener every time the dropdown is open, and if you click somewhere outside of the div where dropdownRef is added, it will close. 

#1166 is fixed by storing the DOM nodes from mouseDown and mouseUp, and since `stopPropagation()` was removed, you'll also want to check if the user is clicking on the Scrim or the Sidesheet.

### DIFF
--- a/src/components/general/Dropdown/index.tsx
+++ b/src/components/general/Dropdown/index.tsx
@@ -2,7 +2,6 @@ import { useState, useRef, useCallback, FC, useEffect, MutableRefObject, ReactNo
 
 import classNames from 'classnames';
 import {
-    useClickOutsideOverlayPortal,
     RelativeOverlayPortal,
     useElevationClassName,
     useRelativePositioning,
@@ -68,7 +67,6 @@ const Dropdown: FC<DropdownProps> = ({ controller, justification, children }) =>
     const { isOpen, setIsOpen, node, controllerRef } = controller;
 
     const close = useCallback(() => isOpen && setIsOpen(false), [isOpen]);
-    useClickOutsideOverlayPortal(close, controllerRef.current);
 
     const dropdownContainerClassNames = classNames(
         styles.dropdownContainer,
@@ -105,6 +103,24 @@ const Dropdown: FC<DropdownProps> = ({ controller, justification, children }) =>
         setMaxHeight(`calc(100vh - ${dropdownRect.top}px - (var(--grid-unit) * 3px))`);
     }, [rect, isOpen]);
 
+    const handleClick = (e) => {
+        if (dropdownRef.current.contains(e.target)) {
+            return;
+        }
+        close();
+    };
+
+    useEffect(() => {
+        if (isOpen) {
+            document.addEventListener('click', handleClick);
+        } else {
+            document.removeEventListener('click', handleClick);
+        }
+
+        return () => {
+            document.removeEventListener('click', handleClick);
+        };
+    }, [isOpen]);
     return (
         <>
             {node}
--- a/src/components/general/Scrim/index.tsx
+++ b/src/components/general/Scrim/index.tsx
@@ -1,6 +1,6 @@
 import classNames from 'classnames';
 import styles from './styles.less';
-import { FC, ReactNode, MouseEvent } from 'react';
+import { FC, ReactNode, MouseEvent, useState, useRef } from 'react';
 
 type ScrimProps = {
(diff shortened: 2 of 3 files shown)


src/components/general/SearchableDropdown/index.tsx
```diff
@@ -208,7 +208,11 @@ const SearchableDropdown = ({
                         placeholder={placeholder || 'Type to search...'}
                         label={label}
                         icon={<DropdownArrow cursor="pointer" isOpen={isOpen} />}
-                        onIconAction={() => isOpen && setIsOpen(false)}
+                        onIconAction={(e) => {
+                            e.stopPropagation();
+                            setIsOpen(!isOpen);
+                            ref.current && ref.current.click && ref.current.click();
```
jaysencpp, line 214: There's a problem with `onIconAction` and `onClick` because a click on the icon will also cause the `onClick` to fire off which results in not being able to close the dropdown when clicking the icon. 
This might be a hacky solution, but by stopping event bubbling from the icon, the `onClick` will not run, however stopping event bubbling will also cause a problem with the click outside logic inside `Dropdown`. It can be worked around by calling on the `click()` method from the ref from `useDropdownController`-hook.

src/components/general/SearchableDropdown/index.tsx
```diff
@@ -186,7 +186,7 @@ const SearchableDropdown = ({
         const overlayContainer = useOverlayContainer();
         const handleBlur = useCallback(
             (e: FocusEvent<HTMLInputElement>) => {
-                if (overlayContainer.contains(e.relatedTarget as HTMLElement)) return;
+                if (!overlayContainer.contains(e.relatedTarget as HTMLElement)) return;
```
jaysencpp, line 189: Not sure about this one. I negated the if-statement because `overlayContainer` will always contain `e.relatedTarget`. 
After negating it, you can use Tab from one SearchableDropdown to another, and the former will close.
overlayContainer (which comes from  `const overlayContainer = useOverlayContainer();`) should maybe be replaced with something else

src/components/general/Dropdown/index.tsx
```diff
@@ -105,8 +112,26 @@ const Dropdown: FC<DropdownProps> = ({ controller, justification, children }) =>
...
+        if (isOpen) {
+            document.addEventListener('click', handleClick);
+        } else {
+            document.removeEventListener('click', handleClick);
+        }
+
+        return () => {
+            document.removeEventListener('click', handleClick);
+        };
+    }, [isOpen]);
```
odinr, line 132: wrap handle click inside useCallback and add the fn as dep to effect

src/components/general/Scrim/index.tsx
```diff
@@ -9,12 +9,30 @@ type ScrimProps = {
...
 const Scrim: FC<ScrimProps> = ({ children, show, onClick }) => {
+    const [mouseDownTarget, setMouseDownTarget] = useState({});
+    const [mouseUpTarget, setMouseUpTarget] = useState({});
+
+    const ref = useRef<HTMLDivElement | null>(null);
     const scrimClassNames = classNames(styles.scrim, {
         [styles.show]: show,
     });
 
+    const handleClick = (e: MouseEvent<HTMLDivElement>) => {
```
odinr, line 20: missing useCallback

src/components/general/SearchableDropdown/index.tsx
```diff
@@ -208,7 +208,11 @@ const SearchableDropdown = ({
                         placeholder={placeholder || 'Type to search...'}
                         label={label}
                         icon={<DropdownArrow cursor="pointer" isOpen={isOpen} />}
-                        onIconAction={() => isOpen && setIsOpen(false)}
+                        onIconAction={(e) => {
```
odinr, line 211: could you assign this to a useCallback

odinr (changes requested):
test

odinr (approved):

PR #1318, opened 2021-11-01
fix: Content style update

AbdullZahir:
changed from less to react styles in fusion-components Content


src/components/core/Content/Content.tsx
```diff
@@ -1,10 +1,12 @@
 import { FC } from 'react';
-import styles from './styles.less';
+//import styles from './styles.less';
```
eskil0312, line 2: remove comment

PR #1319, opened 2021-11-01
fix: Context selector style update

AbdullZahir:
Removed styles.less file with ts files for contextSelector component


odinr (approved):

src/components/core/ContextSelector/ContextSelector.tsx
```diff
@@ -19,7 +19,8 @@ import {
     IconButton,
     CloseIcon,
 } from '@equinor/fusion-components';
-import styles from './styles.less';
+//import styles from './styles.less';
```
eskil0312, line 22: remove comment

PR #1320, opened 2021-11-02
add id to dialog container and the buttons

QiJin-Bouvet:


QiJin-Bouvet:
This PR need PR 1311 as dependency.

odinr (approved):

PR #1325, opened 2021-11-02
feat: bookmarksmanager with more control through props

jardt:
New optional props for more control:

- onBookmarkDeleted:  callback for deleted bookmark
- onBookmarkEdited: callback for edited bookmark
- openOnCreate: boolean to decide if flyout is opened in All bookmarks view mode or create mode
- disableCreateButton: boolean used to disable add bookmarks button

We also want to be able to control open / closed state of the bookmarks sidesheet.
To do this we need to have this state in the app.
The solution proposed here renames the old index file to BookmarksManager which has the props to control the sidesheet state, and exports its.
The new index file is a "wrapper" containing the state for the sidesheet so by importing this the component works as before if apps dont want to control the state.


odinr (approved):

jaysencpp:
Could you fix the merge conflict, @jardt

PR #1327, opened 2021-11-03
PowerBI -  use new version of bookmarks manager

jaysencpp:
Using the new version of Bookmarks Manager component. Related to issue #1330 
Needs to be tested with how the new PowerBI component works (edit: tested it, works fine)

Couple of things:
* AppSettingsManager is deprecated, use BookmarksManager instead
* The `component` ref initialized in the PowerBI provider would always be undefined. It's fixed now by changing the `getEmbeddedComponent` logic. 
* The `contextRef` prop passed to the PowerBI component  would also not be assigned the correct value.
* Using refs in dependency arrays has no effect, so removed them

### DIFF
--- a/src/components/data/PowerBI/Report/components/view/PowerBIReportView.tsx
+++ b/src/components/data/PowerBI/Report/components/view/PowerBIReportView.tsx
@@ -1,7 +1,5 @@
-import { useContext, useEffect, FC, useCallback, useMemo } from 'react';
-
+import { useContext, useEffect, FC, useMemo } from 'react';
 import { PowerBIEmbed, EventHandler } from 'powerbi-client-react';
-
 import { context, PowerBIEmbedEvents } from '../../context';
 import { IEmbedConfiguration, service as PowerBIServices, factories } from 'powerbi-client';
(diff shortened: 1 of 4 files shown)


src/components/data/PowerBI/Report/components/view/PowerBIReportView.tsx
```diff
@@ -47,10 +45,9 @@ export const PowerBIReportView: FC<PowerBIComponentProps> = ({ config }: PowerBI
         [event$]
     );
 
-    const getEmbeddedComponent = useCallback(
-        (value) => (component?.current ? (component.current = value) : null),
-        [component]
-    );
+    const getEmbeddedComponent = (value) => {
+        component.current = value;
```
jaysencpp, line 49: Seems to be some type error here? `getEmbeddedComponent` should receive `value: Embed` but then cannot assign `component.current` to valuje because it's of type `PowerBIEmbedComponent`. Still works though..

olerichard (approved):

odinr (approved):

PR #1343, opened 2021-11-08
Accordion style update

AbdullZahir:
added styling in .ts instead of styles.less. The only missing is to include the .grid-unit module in the .ts file.


eskil0312 (approved):

eskil0312:
is the .grid-unit module added

PR #1354, opened 2021-11-15
updating styling for SideSheet component

AbdullZahir:

PR #1355, opened 2021-11-16
fix: Sidesheet - Remove border from collapse button 

olerichard:
Remove border-bottom from .collpaseButtonContainer in the sidesheet component.

This added a border under the collapse button when using resizable sidesheet.
But the header container in general does not have an underline. Making this look like a short line sticking out into nothing.


odinr (approved):

olerichard:
[updating styling for SideSheet component #1354](https://github.com/equinor/fusion-components/pull/1354) was updating the CSS on this component to use new style framework. 

Asked Abdullah to add this fix there instead so it is not lost. 
Closing this PR.

PR #1411, opened 2021-12-23
Position picker style update

AbdullZahir:


src/components/org/PositionPicker/components/ItemComponent.tsx
```diff
@@ -1,10 +1,11 @@
 import ItemComponentProps from './itemComponentProps';
-import styles from '../styles.less';
+import { useStyles } from '../style';
```
eskil0312, line 2: we should try to be consistent with the styling files, `itemComponent.styles.ts`?

src/components/org/PositionPicker/style.ts
```diff
@@ -0,0 +1,18 @@
...
+            marginLeft: 'calc(var(--grid-unit) * 1px)',
+        },
+        positionName: {
+            fontSize: '14px',
+            lineHeight: '24px',
+        },
+        personName: {
+            fontSize: '14px',
+            lineHeight: '24px',
+        },
```
eskil0312, line 15: name?

PR #1440, opened 2022-01-27
Scrim style update

AbdullZahir:


src/components/general/Scrim/Scrim.style.ts
```diff
@@ -0,0 +1,21 @@
+import { createStyles, makeStyles } from '@equinor/fusion-react-styles';
+
+export const useStyles = makeStyles(
```
eskil0312, line 3: When we export the styles, could we name the hook to be more spesific to its use instead if `useStyles`. When naming all stylig hooks `useStyles`, you will get x amount of results in intellij when importing the file. 
How about `useScrimStyles`?

eskil0312 (approved):

PR #1493, opened 2022-05-06
Add options for style

thomasFjorstad:
Added opportunity to create own table styles

### DIFF
--- a/src/components/data/DataTable/__stories__/DataTable.stories.tsx
+++ b/src/components/data/DataTable/__stories__/DataTable.stories.tsx
@@ -121,6 +121,51 @@ const WithoutSkeleton: FC = () => {
         />
     );
 };
+const DefaultWithOwnStyle: FC = () => {
+    const [appSettings, setAppSetting] = useAppSettings();
+    const perPage = parseInt(appSettings['perPage'], 10) || 20;
+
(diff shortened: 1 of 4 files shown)


src/components/data/DataTable/__stories__/DataTable.stories.tsx
```diff
@@ -121,6 +121,51 @@ const WithoutSkeleton: FC = () => {
...
+    const { sortedData, setSortBy, sortBy, direction } = useSorting(data, null, null);
+    const { pagination, pagedData, setCurrentPage } = usePagination(
+        sortedData as DataItem[],
+        perPage
+    );
+
+    const onPaginationChange = useCallback((newPage: Page, perPage: number) => {
+        setCurrentPage(newPage.index, perPage);
+        setAppSetting('perPage', perPage);
+    }, []);
```
AbdullZahir, line 137: Should'nt the useCallback have dependencies or is it ment to not have any depencies that updates the funtion?

AbdullZahir (changes requested):

src/components/data/DataTable/__stories__/DataTable.stories.tsx
thomasFjorstad: Dep on func?
I just copy the other func in this story file.

AbdullZahir (approved):

PR #1537, opened 2022-09-30
Fix PersonCard re-render issue

joakim-equinor:
Issue: _PersonCard_ does not re-render if you first provide an invalid _personId_, then later provide a valid _personId_.

Solution: Reset the error state when executing the try block.


joakim-equinor:
Will create a new pull request using conventional commits.

PR #1592, opened 2023-02-02
Fix: checking for invalid interval for overlapping dates

jardt:
Overlapping indication on timline preview for position splits can have from date after to date and vice verca.
This invalid interval will cause error. This is a fix for this.


tony-hagen (approved):
LGTM

PR #1596, opened 2023-02-20
Bugfix/navigation menu bugs

tony-hagen:
Sections that does not have a navigation action connected to it will now show/hide children on click of the whole box and not just the arrow. 
Navigation popover no longer have hover and clickable styling on sections if the section does not navigate.

https://statoil-proview.visualstudio.com/Fusion/_workitems/edit/33374


jardt (approved):
Good 💩

Issue #1606, opened 2023-02-22
NavigationDrawer is broken in version 2.10.1


v3gard:
With the newest version, fusion-components `2.10.1`, it seems that `<NavigationDrawer/>` is broken. When I click a row in the component, the `onChangeSelectedId` event doesn't fire, and thus doesn't invoke my callback function.

We typically use `NavigationDrawer` like this. This code has seen no changes for the past 20 months in our repository:

```ts
<Fusion.NavigationDrawer
    id="navigation-drawer-story"
    structure={props.productGroupStructure}
    selectedId={selectedId}
    onChangeSelectedId={props.onProductGroupChanged}
    onChangeStructure={() => {
        /* Empty because we control structure change from the parent function */
    }}
/>
```

The version we're currently using, fusion-components `2.9.3`, seems to work. Perhaps #1596 is the PR that introduced the bug?

# Demo
|Before (2.9.3)|After (2.10.1)|
|-|-|
|![before](https://user-images.githubusercontent.com/343078/220575581-b6a531c2-9b56-4f0d-a1d8-fe2f2022015d.gif)|![after](https://user-images.githubusercontent.com/343078/220576220-54512b02-471a-4b87-ac78-2e637592b257.gif)|

v3gard:
This issue is still present in the [newest release](https://www.npmjs.com/package/@equinor/fusion-components) on npmjs, `2.10.3`. The newest version here on Github appears to be `2.10.2`. This is a bit confusing.

|Github|npmjs|
|-|-|
|![image](https://user-images.githubusercontent.com/343078/221879143-44d2c2c6-53c8-4b44-b79f-24cd4fb193c7.png)|![image](https://user-images.githubusercontent.com/343078/221879385-ce266411-32c4-4b40-8dea-c8b1fd2ba78a.png)|

v3gard:
NavigationDrawer is still broken in 2.10.5.

vft-equinor:
Still an issue in `2.10.6`.

v3gard:
We got a warning from Dependabot today. We cannot solve[ a security vulnerability in the yaml package](https://security.snyk.io/vuln/SNYK-JS-YAML-5458867) because we're required to use an older version of fusion-components due to the issue mentioned above.

![image](https://user-images.githubusercontent.com/343078/234195344-a722c3d9-f359-4b71-8e4e-eb057e236571.png)

Probably something to discuss on the next architecture meeting.

vft-equinor:
[...]

PR #1610, opened 2023-02-23
NavigationDrawer - Hover and onClick improvements

tony-hagen:
Sections and groupings without a navigation action should not be hoverable in the navigation popover.
If navigation drawer is not collapsed, sections and groupings without a navigation action should show/hide children when clicking anywhere on the item, not just the arrow.


eikeland (approved):

eikeland:
Related issue #1606.

